PR #39, opened 2021-11-11
Update namespace. Add table_name config to all models

chrispenny:
Prefixed each with `Link_` to make it easier to find all relevant tables, but also because `FileLink` is already used by framework.


src/Models/EmailLink.php
```diff
@@ -12,6 +12,8 @@
 class EmailLink extends Link
 {
 
+    private static $table_name = 'Link_EmailLink';
```
mfendeksilverstripe, line 15:
```suggestion
    private static $table_name = 'LinkFieldLink_EmailLink';
```

src/Models/ExternalLink.php
```diff
@@ -10,11 +10,12 @@
 class ExternalLink extends Link
 {
 
+    private static $table_name = 'Link_ExternalLink';
```
mfendeksilverstripe, line 13:
```suggestion
    private static $table_name = 'LinkFieldLink_ExternalLink';
```

src/Models/FileLink.php
```diff
@@ -8,17 +8,19 @@
...
 /**
  * A link to a File track in asset-admin
+ *
  * @property File $File
  * @property int $FileID
  */
 class FileLink extends Link
 {
 
+    private static $table_name = 'Link_FileLink';
```
mfendeksilverstripe, line 18:
```suggestion
    private static $table_name = 'LinkFieldLink_FileLink';
```

src/Models/Link.php
```diff
@@ -21,6 +21,8 @@
 class Link extends DataObject implements JsonData, Type
 {
 
+    private static $table_name = 'Link';
```
mfendeksilverstripe, line 24:
```suggestion
    private static $table_name = 'LinkFieldLink';
```

src/Models/SiteTreeLink.php
```diff
@@ -16,6 +16,8 @@
 class SiteTreeLink extends Link
 {
 
+    private static $table_name = 'Link_SiteTreeLink';
```
mfendeksilverstripe, line 19:
```suggestion
    private static $table_name = 'LinkFieldLink_SiteTreeLink';
```

mfendeksilverstripe (changes requested):
I think we should go with a less generic table names to prevent potential naming conflicts. I can definitely see table called `Link` having a conflict especially if a project is migrating from another module. I added `LinkFieldLink` prefix as the name of this module is `LinkField`.

mfendeksilverstripe (changes requested):
@chrispenny The changes look good but it seems some of the functionality is broken because of this change. This is my test setup (TNZ core after updating namespace references):

* Block page with a TextBlock / ButtonLink
* I switcher the block to NOT be inline editable
* navigated to the block edit form

Expected:

Screen Shot 2021-11-16 at 8 45 37 AM

Actual:

Screen Shot 2021-11-16 at 8 42 20 AM

Maybe there are some class references in the React bundle as well?

mfendeksilverstripe:
@chrispenny Can you please check if you are able to save and retain link related changes? I can see that the UI now works but none of the changes I make are retained. Seems to be an issue on both inline edit and normal edit form cases.

Create a link

Screen Shot 2021-11-16 at 11 41 11 AM

Hit save

Screen Shot 2021-11-16 at 11 41 20 AM

Link doesn't show up even after page reload.

chrispenny:
Hi @mfendeksilverstripe,

These seem to be working ok for me.

Inline editor (text block):
Screen Shot 2021-11-16 at 1 29 23 PM

Standard editor (cta block), straight after a publish:
Screen Shot 2021-11-16 at 1 30 49 PM

Standard editor (cta block), after a refresh:
Screen Shot 2021-11-16 at 1 31 01 PM

I don't believe I have done anything special other than a `dev/build` and appropriate flushes.

mfendeksilverstripe (approved):
Tested on local and it's working fine 👍 
thanks for contribution :)

mfendeksilverstripe:
@chrispenny Feel free to merge.

PR #47, opened 2022-06-28
PHP 8.1: Suppress deprecation notice

chrispenny:
`JsonSerializable` has defined the return type of `:mixed` for the `jsonSerialize()` method. I don't think we can define this in our module until we support a minimum of PHP 8.1.


mfendeksilverstripe (approved):
LGTM 👍

PR #48, opened 2022-07-14
Add PhoneField. Add migration helpers from Linkable

chrispenny:
* Linkable had a concept of a "Phone" link, which didn't currently exist in LinkField. So I've added that. It's really just a duplicate of `EmailField`, but it returns a `tel` instead of a `mailto`.
* Added some instructions based on what I found from migrating my own project. These steps may not be completely comprehensive, but I've tried to at least indicate areas that might need attention.
* Supports `Versioned` and `un-Versioned`. However, we expect the `Versioned` state between legacy (`Linkable`) and `LinkField` to match.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,3 +57,12 @@ class Page extends SiteTree
     }
 }
 ```
+
+## Migrating from Shae Dawson's Linkable module
+
+https://github.com/sheadawson/silverstripe-linkable
+
+Shae Dawson's Linkable module was a much loved, and much used module. It is, unfortunately, no longer maintained. We
+have provided some steps and tasks that we hope can be used to migrate your project from Linkable to LinkField.
+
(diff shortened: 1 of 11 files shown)


GuySartorelli:
The migration stuff feels quite separate to the phone link functionality. Might be worth splitting it up into two PRs?

mfendeksilverstripe (approved):
Awesome work 👍 Code is looking good, documentation is quite detailed and it works for me locally. I managed to migrate a bunch of links from my fixture.

Screen Shot 2022-10-28 at 8 17 40 AM

I added some very minor changes:

* WS fixes
* added a few doc-blocks
* renamed a few variables (avoid type in variable name)
* simplified a few DB method calls
* opened up some properties / methods to be protected (more flexibility in case someone needs to subclass)

PR #45, opened 2022-05-25
Add multi link support

maxime-rainville:
This PRs add support for multi link field via an HasMany relation to Links.

It also adds some extra features like:
- Link Types can now have icons
- Link titles can default to picking up some bit of metadata on their target (e.g: Linking to a CMS page can default to the page title)

# Parent issue
- #44

PR #54, opened 2023-02-27
v2: Update namespace, dependencies. Add PhoneLink & migration task

chrispenny:
## Updating `master` (GraphQL 4) with important changes in `1` (GraphQL 3)

I think for a long time development was only being undertaken for the `1` branch, so `master` (which was effectively `2`) has fallen quite far behind.

* Update Github Actions and Recipe Testing
* Update namespace from `Link` to `LinkField`.
* Bring over the `PhoneLink` model.
* Bring over the `LinkableMigrationTask`.
* Add missing `table_name` configuration.
* `SiteTreeLink` should grab `Title` from its page, if no title is specified.
* Linting

**What I didn't bring over**

* Client file changes (Mojmir indicated that these were just linting changes, I think bringing these over will make #45 more difficult to rebase).

## Testing

* `Page` `has_one` `Link::class`: tested within the page edit form.
* `Page` `has_many` `LinkItem`, which `has_one` `Link`: tested within a GridField edit form.

### Issues with GridField edit form

Currently the scaffolded field seems to be a dropdown of existing `Link` models, rather than a `LinkField`. I'm not sure why that is.

Screen Shot 2023-02-28 at 9 23 33 AM

I think this issue exists in the GraphQL 3 version as well, so probably doesn't have to be solved here.

The workaround is to remove the scaffolded field, and explicitly add a `LinkField`.


.github/workflows/main.yml
```diff
@@ -0,0 +1,12 @@
...
+on:
+  push:
+  pull_request:
+  workflow_dispatch:
+
+jobs:
+  ci:
+    uses: silverstripe/gha-ci/.github/workflows/ci.yml@v1
+    with:
+      js: false
```
mfendeksilverstripe, line 12: This should not be disabled. The linting changes from `1` don't need to be carried over but we should have a JS pipeline running as it's considered part of the Github actions setup. My suggestion is to:

* enable JS pipeline
* carry over a sample JS test so we know the pipeline is running
* skip the linting changes to avoid merge conflicts

phpunit.xml.dist
```diff
@@ -1,10 +1,16 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!--
-Standard module phpunit configuration.
-Requires PHPUnit ^5.7
--->
 <phpunit bootstrap="vendor/silverstripe/framework/tests/bootstrap.php" colors="true">
-	<testsuite name="Default">
-		<directory>tests/php</directory>
-    </testsuite>
+    <testsuites>
```
mfendeksilverstripe, line 3: Looks like you haven't carried over schema changes from `1`.

```
onfiguration: /home/runner/work/silverstripe-linkfield/silverstripe-linkfield/phpunit.xml.dist
Warning:       Your XML configuration validates against a deprecated schema.
Suggestion:    Migrate your XML configuration using "--migrate-configuration"!
```

mfendeksilverstripe (changes requested):
Looking good overall 👍 , found a couple of minor issues though.

src/Tasks/LinkableMigrationTask.php
```diff
@@ -0,0 +1,434 @@
...
+/**
+ * @codeCoverageIgnore
+ */
+class LinkableMigrationTask extends BuildTask
+{
+    protected const TABLE_BASE = 'LinkableLink';
+    protected const TABLE_LIVE = 'LinkableLink_Live';
+    protected const TABLE_VERSIONS = 'LinkableLink_Versions';
+
+    protected const TABLE_MAP_LINK = [
```
maxime-rainville, line 28: FYI We could get those table names with something like `DataObject::getSchema()->tableName(Link::class);`

That way, if some mad scientist decide to change the table name for their models, your script will still work.

src/Models/Link.php
```diff
@@ -1,31 +1,44 @@
...
- * A Link Data Object. This class should be subclass and you should never directly interact with a plain Link instance.
+ * A Link Data Object. This class should be a subclass, and you should never directly interact with a plain Link
+ * instance
  *
  * @property string $Title
  * @property bool $OpenInNew
  */
 class Link extends DataObject implements JsonData, Type
 {
+    private static $table_name = 'LinkField_Link';
```
maxime-rainville, line 29: This is definitively an improvement. However, when I run a `dev/build`, pre-existing data doesn't get moved to the new table. Do we have solution for that? Do we need to call this out somewhere for people upgrading form 1 to 2?

maxime-rainville (changes requested):
Did a quick code review and didn't see anything obviously wrong. I'm assuming that all of this code was more or less lifted from the v1 version of the module.

I tested that the core functionality of the module still works. I tested by adding a link to a page and by creating a Link block.

Didn't test the migration ... I don't have much context on how that other module work.

My main concern is what will happen with the data with the new table name?

src/Models/Link.php
chrispenny: @maxime-rainville ^1.1 has the same table names:
https://github.com/silverstripe/silverstripe-linkfield/blob/1/src/Models/Link.php#L29

The issue is probably more for folks going from `dev-master`/`2.x-dev` to our official tag. I can certainly add some comments regarding this though. I'm not sure I want to go as far as to provide a migration task - since `dev-master` was flagged as experimental.

.github/workflows/main.yml
chrispenny: * Enabled JS testing
* Ignore all JS linting
* Brought across `sample-test.js`
* Updated dist files (apparently that breaks js tests if there is a `diff` in the build)

https://github.com/silverstripe/silverstripe-linkfield/pull/54/commits/71c50537d1e7299d39105453efbbd68b38e2f94b

src/Tasks/LinkableMigrationTask.php
chrispenny: Update here 😄 Thanks again!
https://github.com/silverstripe/silverstripe-linkfield/pull/54/commits/be7ece020e7c621410029409bc3c06ac32bbf18d

.eslintignore
```diff
@@ -1,2 +1 @@
-client/dist/
-client/lang/
+client/
```
mfendeksilverstripe, line 1: This effectively disables linting on the frontend files. I suggest to add a eslint ignore (which impacts the whole file) on top of each file instead with a comment that this should be re-enabled at some point. We can then reenable and fix file on case by case basis outside of scope of this PR.

mfendeksilverstripe (changes requested):
Thanks for the changes 👍 I have one minor concern with the frontend assets linting but other than that it's looking really good.

mfendeksilverstripe (approved):
Awesome work 👍 I'm happy with the changes.

src/Models/Link.php
maxime-rainville: That's perfect. I'm happy with that.

maxime-rainville (approved):
Me happy!

PR #55, opened 2023-03-02
Update support for Silverstripe 5

chrispenny:
## SiteTreeLink no longer saves `Page::$Title` to `Link::$Title`

When we're using a `SiteTreeLink`, then it's quite common for us to want to inherit the `Title` of the linked `Page`. Current behaviour is that when you create your `Link`, the `Page::$Title` is saved into the `Link::$Title`. This means that if the `Page::$Title` changes later down the line, your `Link` will still have whatever the original `Title` was.

Updated logic:

* Added a `getTitle()` method to `SiteTreeLink`, which will first return `Link::$Title`, and fall back to `Page::$Title` (if possible).
* Removed the `onBeforeWrite()`, so we're no longer saving the `Page::$Title` in our `Link` model.
* Added `TitleRelField` to `Link::jsonSerialize()`, and updated the JS component to display this as a fall back if no `Title` was provided.

The outcome is...

In your Link preview, you can see the Title from `TitleRelField`:
Screen Shot 2023-03-29 at 10 13 10 AM

But in the Edit form, the Title remains empty:
Screen Shot 2023-03-29 at 10 13 03 AM


chrispenny:
We've been running this branch on a Silverstripe 5 project for a while now, and haven't had any specific issues. I'll merge this and tag it as `3.0.0-beta1`.

PR #56, opened 2023-03-03
SS5 front-end update

blueo:
updates to new webpack config
update dependencies
removed some unused dependences (i'm sure there are still some more)
tidy the files up a little


webpack.config.js
```diff
@@ -1,16 +1,11 @@
...
 const {
-  resolveJS,
-  externalJS,
-  moduleJS,
-  pluginJS,
-  moduleCSS,
-  pluginCSS,
+  CssWebpackConfig,
+  JavascriptWebpackConfig
 } = webpackConfig;
```
GuySartorelli, line 7: Just do `const { CssWebpackConfig, JavascriptWebpackConfig } = require('copy-webpack-plugin');` - though that's a nitpick, not a big deal if left as is

webpack.config.js
```diff
@@ -1,16 +1,11 @@
...
 
-const ENV = process.env.NODE_ENV;
 const PATHS = {
```
GuySartorelli, line 9: For the paths, you only need to keep `ROOT`, any that are non-standard paths (which is none of these) and any that are used below explicitly (`SRC` and `DIST` only). See https://github.com/silverstripe/webpack-config/tree/2#usage

package.json
```diff
@@ -14,10 +14,10 @@
   },
   "author": "SilverStripe Ltd",
   "engines": {
-    "node": "^10.x"
+    "node": "^18.x"
   },
   "scripts": {
-    "build": "yarn && yarn lint && yarn test && NODE_ENV=production webpack -p --bail --progress",
+    "build": "yarn && yarn lint && yarn test && NODE_ENV=production webpack --mode production --bail --progress",
```
GuySartorelli, line 20:
```suggestion
    "build": "yarn && yarn lint && yarn test && rm -rf client/dist/* && NODE_ENV=production webpack --mode production --bail --progress",
```
Remove everything in dist before we start, so we know we don't have any extra garbage.

package.json
```diff
@@ -48,58 +48,45 @@
     }
   },
   "devDependencies": {
-    "@silverstripe/eslint-config": "0.0.5",
-    "@silverstripe/webpack-config": "^1.5.0",
+    "@babel/runtime": "^7.20.0",
+    "@silverstripe/eslint-config": "1.0.0-alpha6",
```
GuySartorelli, line 52:
```suggestion
    "@silverstripe/eslint-config": "^1.0.0-alpha6",
```
We want to make sure we get updates

package.json
```diff
@@ -48,58 +48,45 @@
...
+    "@silverstripe/webpack-config": "^2.0.0-alpha9",
     "@storybook/addon-actions": "^3.4.11",
     "@storybook/addons": "^3.4.11",
     "@storybook/react": "^3.4.11",
```
GuySartorelli, line 56: Is there any storybook building in this module? Probably remove these.

package.json
```diff
@@ -48,58 +48,45 @@
...
-    "babel-jest": "^23.6.0",
+    "babel-jest": "^29.2.2",
     "copy-webpack-plugin": "^4",
```
GuySartorelli, line 58: copy plugin isn't used, remove it

package.json
```diff
@@ -48,58 +48,45 @@
...
-    "react-router-dom": "^4.4.0-beta.6",
-    "react-selectable": "^2.0.2",
-    "reactstrap": "^6.4.0",
-    "redux": "^4.0.0",
-    "redux-form": "^7.4.2",
-    "redux-logger": "^2.4.0",
-    "redux-thunk": "^1.0.3",
-    "webpack-sources": "^1.1.0"
+    "react-dom": "^18.2.0",
+    "react-load-script": "^0.0.6",
```
GuySartorelli, line 75: react-load-script is abandoned https://www.npmjs.com/package/react-load-script

package.json
```diff
@@ -48,58 +48,45 @@
...
+    "react-redux": "^8.0.4",
+    "react-router": "^6.4.2",
+    "react-router-dom": "^6.4.2",
+    "react-select": "^5.5.8",
+    "reactstrap": "^8.9.0",
+    "redux": "^4.2.0",
+    "redux-form": "^8.3.8",
+    "redux-thunk": "^2.4.1",
+    "webpack-sources": "^3.2.3"
   },
```
GuySartorelli, line 85:
```suggestion
  },
  "browserslist": [
    "defaults"
  ],
```
Make sure we're explicitly stating the browserset we support here

package.json
```diff
@@ -48,58 +48,45 @@
...
     "presets": [
-      "env",
-      "react"
-    ],
-    "plugins": [
-      "transform-object-rest-spread"
+      "@babel/preset-env",
+      "@babel/preset-react"
     ]
   }
```
GuySartorelli, line 91: For non-admin modules it's preferred to use a `babel.config.json` file to ensure jest is able to parse the admin imports correctly. See https://github.com/silverstripe/silverstripe-asset-admin/pull/1330/files#r1100908509
Looks like that's not a problem in this module for whatever reason so prooobably doesn't matter here, but just an FYI

client/src/styles/bundle.scss
```diff
@@ -1,5 +1,4 @@
 // Variables
-@import "link-variables";
```
GuySartorelli, line 2: Why is this removed?

client/src/components/LinkField/LinkField.js
```diff
@@ -27,24 +26,29 @@ const LinkField = ({id, loading, Loading, data, LinkPicker, onChange, types, lin
...
+  const onModalSubmit = (modalData, action, submitFn) => {
+    const { SecurityID, action_insert: actionInsert, ...value } = modalData;
+
+    if (typeof onChange !== 'function') {
+      return Promise.resolve();
+    }
+
+    onChange(event, { id, value });
     setEditing(false);
     setNewTypeKey('');
```
GuySartorelli, line 48:
```suggestion
    if (typeof onChange === 'function') {
      onChange(event, { id, value });
      setEditing(false);
      setNewTypeKey('');
    }
```
This is subjective though so take it or leave it

GuySartorelli: Also... are these lines even related to whether we have an onChange callback or not?
```js
setEditing(false);
setNewTypeKey('');
```

They feel separate.

client/src/components/LinkField/LinkField.js
```diff
@@ -78,6 +82,5 @@ export default compose(
   injectGraphql('readLinkTypes'),
   stringifyData,
   injectGraphql('readLinkDescription'),
-  withApollo,
```
GuySartorelli, line 81: Instead of getting rid of this entirely, import it as `import { withApollo } from '@apollo/client/react/hoc';`
(forgot about this when you asked on slack but seeing it in context with code got my memory pumping)

client/src/bundles/bundle.js
```diff
@@ -6,4 +6,4 @@
 // require('expose-loader?InsertMediaModal!containers/InsertMediaModal/InsertMediaModal');
 
 require('boot');
-require('entwine/JsonField.js');
+require('entwine/JsonField');
```
GuySartorelli, line 9: Consider changing these to `import` instead.

GuySartorelli (changes requested):

GuySartorelli:
Note, I haven't actually run this - these suggestions are based on a visual inspection of the code only.

client/src/components/LinkField/LinkField.js
blueo: yeah though it doesn't seem necessary? The component doesn't access the client instance from what I can see.

client/src/components/LinkField/LinkField.js
blueo: yeah good point - was mainly fixing a linting error will move these up

client/src/styles/bundle.scss
blueo: there was nothing in it so I figured it was unnecessary

client/src/components/LinkField/LinkField.js
GuySartorelli: Fair enough, if it's not needed it's not needed 👍

client/src/styles/bundle.scss
GuySartorelli: Makes sense

maxime-rainville (changes requested):
Tested by adding a LinkField to a page and to a Content block.

The Page one works fine, but when I try to interact with the LinkField inside a block the modal doesn't show up and it looks stuck in some way. Replicated in FireFox and Chrome.

There's nothing in the JS console log.

[Screencast from 2023-03-03 15-43-04.webm](https://user-images.githubusercontent.com/1168676/222618906-3a881dcb-5fa0-469a-a467-b2263497726f.webm)

maxime-rainville:
After a select a type for my block, I get a never ending list of request to DynamicLink.

image

maxime-rainville:
I created a test repo that adds a bunch of dummy LinkField https://github.com/maxime-rainville/silverstripe-linkfield-tester ... I suspect we'll problay nede somthing like this at some point to automate test.

I created a LinkField in the asset-admin Folder form. That form and the Elemental block form work slightly differently. So when creating a new field type, both context should be tested. The Folder LinkField allows you to select a link type and enter data, but that data is lost once try to save the modal. It behaves that way in CMS4 as well, so it's probably beyond the scope of this card to solve this.

blueo:
> After a select a type for my block, I get a never ending list of request to DynamicLink.
> 
> image

Yeah it seems the form loader mechanism is going into an infinite loop. Possibly because of having a nested form builder - the link field creates a dynamic form which updates the form state which triggers the elemental area form to re-render and on we go...

blueo:
OK so I think i've narrowed this down to how elemental is implemented. It seems that every time the formState is updated (eg by a form builder component being rendered AKA the linkfield) then the ElementalEditor will re render. The culprit appears to be this part https://github.com/silverstripe/silverstripe-elemental/blob/4/client/src/components/ElementEditor/ElementEditor.js#L138

If i replace it with the following it works
```javascript
function mapStateToProps(state, {areaId}) {
  if (!areaId) {
    return {};
  }

  const formNamePattern = loadElementFormStateName(areaId);
  const elementFormState = state.form.formState.element;

  if (!elementFormState) {
    return {};
  }

  const formState = elementFormState[formNamePattern]

  return {formState};
}
```

However this gets rid of the regex matching on the state Key. I couldn't see why this was needed - the original PR (https://github.com/silverstripe/silverstripe-elemental/pull/439) looks like originally had this implementation but then it changed.... anyone able to decipher the git history better than me? @maxime-rainville @GuySartorelli ?

GuySartorelli:
I took the liberty of changing your code snippet to a diff so it's really clear what you're changing there.

Looks like that regex was added in the PR you linked to - specifically in [this commit](https://github.com/silverstripe/silverstripe-elemental/pull/439/commits/d254223c13e22bcd2e2803c5c9c75f02ab66b807#diff-0c1ab3692e107d03408de5200c0d4a446265558530d0f2761f3a9ef1d35c73a6R39)

That said I don't have enough background in this area of the code to tell you anything about what that's doing or why it's there 😅

blueo:
> I took the liberty of changing your code snippet to a diff so it's really clear what you're changing there.
> 
> Looks like that regex _was_ added in the PR you linked to - specifically in [this commit](https://github.com/silverstripe/silverstripe-elemental/pull/439/commits/d254223c13e22bcd2e2803c5c9c75f02ab66b807#diff-0c1ab3692e107d03408de5200c0d4a446265558530d0f2761f3a9ef1d35c73a6R39), and then subsequent commits in the same PR changed how it's used slightly.
> 
> That said I don't have enough background in this area of the code to tell you anything about what that's doing or why it's there 😅

thanks - weirdly the very next commit has changed it - https://github.com/silverstripe/silverstripe-elemental/pull/439/commits/606ccab6d9049c1199d00cf4aff5a9d10dd1203d#diff-642820167278809b93baf09f49a985a9563af65057ca3ab441617f034c238d2eR87 perhaps I'll just make a PR there and see if any Behat tests fail :)

blueo:
the infinite loop issue is fixed by: https://github.com/silverstripe/silverstripe-elemental/pull/1044 so this will need to wait for that to merge

blueo:
@maxime-rainville what issue did you have with the files area? I seem to be able to add a linkfield there:
image
and it persists after saving

GuySartorelli:
The elemental bug fix is released as [5.0.0-beta2](https://github.com/silverstripe/silverstripe-elemental/releases/tag/5.0.0-beta2)

PR #76, opened 2023-08-01
ENH Optomise Link models for AnyField

maxime-rainville:
Partially supersedes https://github.com/silverstripe/silverstripe-linkfield/pull/45

This is meant to optimise the Link models to work with my new [AnyField](https://github.com/maxime-rainville/silverstripe-anyfield) module.

I've set this up so it's SEMVER compliant, but honestly my preference would  be to clean up all the bells and whistle that don't make sense anymore with a dedicated AnyField.

### DIFF
--- a/src/Models/EmailLink.php
+++ b/src/Models/EmailLink.php
@@ -18,6 +18,8 @@ class EmailLink extends Link
         'Email' => 'Varchar(255)',
     ];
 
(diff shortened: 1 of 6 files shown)


GuySartorelli:
Closing at this isn't currently part of the plan. Might be done in a future minor release depending on what happens with anyfield.

PR #83, opened 2023-09-25
ENH Simplify APIs

emteknetnz:

Switching out graphql to a regular controller makes it far easier to see what's going on because we can remove 8 files and only have to add 1 new file to get equivalent functionality


maxime-rainville:
We've already done the re factoring of this bit. No point keeping this around.

Issue #17, opened 2020-11-23
Validation of standard link types


maxime-rainville:
## Story
As a content author, I want LinkField to block me from inputting invalid data.

## Acceptance criteria
- Appropriate logic is applied - e.g. external links must have a URL, page links must have a Page, etc.
  - PageLink should have a page
  - EmailLink should have a valid email address
  - External link should have a valid full URL 
  - Phone link should have a non blank string but not otherwise validate the phone number
  - File link should point to a valid CMS file.
- We use form validation for validating the data.
- The LinkField doc is updated to link to the form validation article in the main doc.

## Exclude
- Displaying the validation error in a front end. https://github.com/silverstripe/silverstripe-linkfield/issues/16

## Guy's PRs
- https://github.com/silverstripe/silverstripe-framework/pull/11123
- https://github.com/silverstripe/developer-docs/pull/454

## Steve's PRs
- https://github.com/silverstripe/silverstripe-framework/pull/11116
- https://github.com/silverstripe/silverstripe-linkfield/pull/85
- https://github.com/silverstripe/developer-docs/pull/450

maxime-rainville:
We have already partially implemented this in the process of re factoring bits of the linkfield.

Maybe this card just become a doc card.

GuySartorelli:
PRs merged

PR #85, opened 2023-09-26
ENH Improved validation

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-linkfield/issues/17

Requires https://github.com/silverstripe/silverstripe-framework/pull/11116 to be merged first before CI works


src/Form/PhoneField.php
```diff
@@ -0,0 +1,62 @@
+<?php
+
+namespace SilverStripe\LinkField\Form;
+
+use SilverStripe\Forms\TextField;
+use SilverStripe\Forms\Validator;
+
+/**
+ * Text input field with validation for correct email format according to RFC 2822.
```
emteknetnz, line 9: Need to update this since it's not an email field

purplespider:
For the `ExternalLink` field validation etc, can I suggest considering using https://github.com/burnbright/silverstripe-externalurlfield? (I do feel that module should really be part of the core, as it would help keep external links and their fields consistent between different modules).

src/Models/Link.php
```diff
@@ -110,6 +113,37 @@ public function getCMSCompositeValidator(): CompositeValidator
...
+     * dataobject using DataObject::setData($data)
+     * Because of this alternate method of saving, the getCMSCompositeValidator() isn't called
+     * so we need to call it manually here by loading it into a temporary Form
+     *
+     * @return ValidationResult
+     */
+    public function validate()
+    {
+        $parentResult = parent::validate();
+        $validator = $this->getCMSCompositeValidator();
```
emteknetnz: On second thought this is the wrong place to be calling getCMSCompositeValidator() - instead we should be calling it alongside the DataObject validation, rather that inside in

lang/en.yml
```diff
@@ -3,6 +3,8 @@ en:
     CREATE_LINK: 'Create link'
     MENUTITLE: 'Link fields'
     UPDATE_LINK: 'Update link'
+  SilverStripe\LinkField\Form\ExternalLinkField:
+    INVALID: 'Please enter a valid URL'
```
ChloeSartorelli:
```suggestion
```
This is in framework now.

src/Models/EmailLink.php
```diff
@@ -55,4 +57,11 @@ public function getMenuTitle(): string
     {
         return _t(__CLASS__ . '.LINKLABEL', 'Link to email address');
     }
+
+    public function getCMSCompositeValidator(): CompositeValidator
+    {
+        $validator = parent::getCMSCompositeValidator();
+        $validator->addValidator(RequiredFields::create(['Email']));
```
ChloeSartorelli, line 64:
```suggestion
        $requiredFieldsValidators = $validator->getValidatorsByType(RequiredFields::class);
        if (empty($requiredFieldsValidators)) {
            $requiredFieldsValidator = RequiredFields::create(['Email']);
        } else {
            $requiredFieldsValidator = $requiredFieldsValidators[array_key_first($requiredFieldsValidators)];
        }
        $validator->addValidator($requiredFieldsValidator);
```
An alternative approach would be to add a `getOrCreateValidatorOfType()` method to `CompositeValidator` (similar to [`FieldList::findOrMakeTab()`](https://api.silverstripe.org/5/SilverStripe/Forms/FieldList.html#method_findOrMakeTab)) - or at least a `getValidatorByType()` (similar to [`GridFieldConfig::getComponentByType()`](https://api.silverstripe.org/5/SilverStripe/Forms/GridField/GridFieldConfig.html#method_getComponentByType))

You don't have to do this, but it does make the validator tidier and easier to work with in subclasses or extensions. I won't block merging if you don't make this change.

If you do make this change, please also make it in the other `getCMSCompositeValidator()` implementations.

src/Models/ExternalLink.php
```diff
@@ -27,8 +30,13 @@ class ExternalLink extends Link
     public function getCMSFields(): FieldList
     {
         $this->beforeUpdateCMSFields(function (FieldList $fields) {
-            $linkField = $fields->dataFieldByName('ExternalUrl');
-            $linkField->setTitle(_t(__CLASS__ . '.EXTERNAL_URL_FIELD', 'External url'));
+            $field = UrlField::create('ExternalUrl');
+            $field->setTitle(_t(__CLASS__ . '.EXTERNAL_URL_FIELD', 'External URL'));
```
ChloeSartorelli, line 34:
```suggestion
            $field = UrlField::create('ExternalUrl', $this->fieldLabel('ExternalUrl'));
```
This should have been done like this originally, since the field label localisation is provided (see `db_ExternalUrl`) - but I won't block merging if you don't make this change since it's not directly related to what this PR is for.

If you do make this change, please make it for other link types as well.

GuySartorelli (changes requested):

lang/en.yml
emteknetnz: Removed

src/Models/EmailLink.php
emteknetnz: Leaving as is

src/Models/ExternalLink.php
emteknetnz: Leave as is

README.md
```diff
@@ -164,6 +164,10 @@ class MyCustomLink extends Link
 }
 ```
 
+## Custom link validation
+
+Custom links can have validation set using standard [model validation]([https://docs.silverstripe.org/en/5/developer_guides/forms/validation/#model-validation).
```
ChloeSartorelli:
```suggestion
Custom links can have validation set using standard [model validation](https://docs.silverstripe.org/en/5/developer_guides/forms/validation/#model-validation).
```

GuySartorelli (approved):
Looks good, will merge after framework PR.

README.md
ChloeSartorelli: Still need to fix this minor thing

GuySartorelli (changes requested):

GuySartorelli:
Works great on a page.

When I try in an inline-elemental block, the first time I try to create a link and trigger validation (i.e. an invalid link), the modal just closes and reopens.
Submitting again triggers validation and gives the message correctly, so it's just the first time submitting the modal form with an invalid link.

README.md
emteknetnz: Updated

emteknetnz: Updated

emteknetnz:
> When I try in an inline-elemental block, the first time I try to create a link and trigger validation (i.e. an invalid link), the modal just closes and reopens. Submitting again triggers validation and gives the message correctly, so it's just the first time submitting the modal form with an invalid link.

I've spent half a day debugging this.  This is seems like an extremely difficult issue to fix.  Core issue is that the ElementList.js (possibly higher up) is re-rendering on validation failure, causing the nested LinkField and nested FormBuilderModal to themselves re-render.

LinkFields on Page don't have this issue because they're not nested inside a react component higher up that keeps re-rendering

This can be validated by putting in the following into (silverstripe/silverstripe-elemental) ElementEditor.js and putting breakpoints in and causing the validation to fail - componentDidUpdate() will trigger on the 1st validation failure. It will not trigger on the 2nd validation failure.

```js
  componentDidMount() {
    var x = 1;
  }

  componentDidUpdate() {
    var x = 1;
  }
```

What's causing the re-render seems like its in the "magical blackbox" territory of JS Injector which does a bunch of stuff in registerTransforms.js and possibly redux-form.  I don't really know.  I've pretty confident that if we're able to get a fix for this it won't be in linkfield, it'll likely be in elemental though possibly admin instead

There's an outside chance there's some cross-over with https://github.com/silverstripe/silverstripe-admin/issues/1662, 

Debugging in entwine.js on `root.render(<ElementEditorComponent {...props} />);` this does NOT trigger a breakpoint the same way as other things do, so ElementEditor.js is the top-most bit

emteknetnz:
Update: Max traced it to this [line](https://github.com/silverstripe/silverstripe-linkfield/blob/03ffff8497fe604828738f0a925ec27a37ffafa7/client/src/containers/LinkModalContainer.js#L18) - making it so that Injector isn't used fixes the issue and instead we just use `<LinkModal>` as per normal fixes issue

I've leave this code as is in this PR and look to fix or remove [on this card](https://github.com/silverstripe/silverstripe-linkfield/issues/202)

GuySartorelli:
For my money, passing the injected modal as a prop or memoising it seems likely to resolve this.

emteknetnz:
I had a quick try with

```js
// https://react.dev/reference/react/memo
import React, { memo } from 'react';

const LinkModal = memo(loadComponent(`LinkModal.${handlerName}`));
```

Didn't work

~~Passing in a prop (haven't tested) is kinda ugly because `<LinkModalContainer>` can come from either LinkField.js or LinkPicker.js~~

Passing as a prop from LinkPicker.js to LinkModalContainer.js didn't seem to work either

GuySartorelli:
Max came up with a solution. I've simplied it a lot, the only part that's really needed is setting the `loadComponent` result as state like so:

```js
import React, { useState } from 'react';

// ...
const [ LinkModal, setLinkModal ] = useState(null);

// ...
if (LinkModal === null) {
  setLinkModal(() => loadComponent(`LinkModal.${handlerName}`));
}
```

You can wrap setting the state in a call to `useEffect` as well if you like (using `[typeKey]` as the second argument) - but in my (admittedly limited) test that didn't seem necessary.

README.md
ChloeSartorelli: I saw that this had been fixed, but now it's unfixed?

GuySartorelli (changes requested):
Works great locally, just a tiny docs regression

README.md
emteknetnz: Updated

GuySartorelli (approved):
LGTM

Issue #84, opened 2023-09-25
Investigate linkfield improvements


emteknetnz:
Holder card for potential linkfield improvement PRs

This implement some of what was talked about in these issues
- https://github.com/silverstripe/silverstripe-linkfield/issues/81
- https://github.com/silverstripe/silverstripe-linkfield/issues/78

I've made them all draft PRs and I've mainly only included code changes and not unit tests or docs, the intention being that we may not want some of the PRs and because there's an existing unmerged PR from Max to add multi-link support which may impact some of the draft PRs

We'll probably want to merge this old has_many PR Max did first as is, fix any broken tests and then rebase the other PRs and then take it for there
- https://github.com/silverstripe/silverstripe-linkfield/pull/45

## PRs
- https://github.com/silverstripe/silverstripe-linkfield/pull/85
- https://github.com/silverstripe/silverstripe-linkfield/pull/88
- https://github.com/silverstripe/silverstripe-linkfield/pull/86
- https://github.com/silverstripe/silverstripe-linkfield/pull/87
- https://github.com/silverstripe/silverstripe-linkfield/pull/90
- https://github.com/silverstripe/silverstripe-linkfield/pull/83
- https://github.com/silverstripe/silverstripe-linkfield/pull/91

From there we'll also want to do the following:
- Refactoring / tidy up
- Ensure adequate test coverage
- New readme
- Add developer docs, including how to add custom links types or modify existing ones
- Maybe look at UI to make it look more like the TinyMCE plugin

#### Other
- https://github.com/silverstripe/silverstripe-linkfield/pull/89

GuySartorelli:
@maxime-rainville @emteknetnz This should not be unassigned

emteknetnz:
Sorry accidentally assigned Max then unassigned him, should have just stayed on me, have reassigned myself)

purplespider:
Linking to #40 re consistency with TinyMCE dropdown.

maxime-rainville:
I've attached all the PR to relevant issues or closed them if they weren't relevant.

PR #86, opened 2023-09-26
NEW Set allowed link types

emteknetnz:

Note this is currently only a soft limit and isn't validated on the server

Will also need to add similar method to MultiLinkField if we merge https://github.com/silverstripe/silverstripe-linkfield/pull/45


maxime-rainville:
Closed in favour of https://github.com/silverstripe/silverstripe-linkfield/pull/137

PR #87, opened 2023-09-26
ENH Allow projects to define their own subclass templates

emteknetnz:

Currently you're stuck having to use either the default `Link.ss` template or overriding all Link templates globally by creating a 'Link.ss' in your project

This allows you to defined templates for individual Link types in your project


maxime-rainville (approved):
This a great idea. I rebased the PR and validated it works locally.

It's a pretty straightforward change. Don't think we need to discuss this any further.

Issue #7, opened 2020-11-23
Versioned by default


maxime-rainville:
## Story
As a **CMS user** I want **the ability to control the publication state of links through it's page (or other parent object)** so that I can view draft version and restore older version.

## Acceptance criteria
- The versioned extension is applied to the Link Object.
- The LinkField UI provide feedback about the publication state of the link
- The need to define ownership is made explicit in the documentation.
- Links can still be used if they are unversioned.
- UX team provides feedback and guidance on how to implement the versioned UI.

## Related
- https://github.com/silverstripe/silverstripe-linkfield/issues/97
- https://github.com/silverstripe/silverstripe-linkfield/issues/10

## Note
- Project will need to use ownership relations are used to publish/unpublish links. It will be the responsibility of individual project to configure that.
- This draft may be useful https://github.com/silverstripe/silverstripe-linkfield/pull/90/files

## Designs
[Link field - Overview](https://projects.invisionapp.com/dsm/silver-stripe/silverstripe-cms/nav/build/folder/components/5fbc28d0096742ec25194c31?mode=preview)

## New issues created
- https://github.com/silverstripe/silverstripe-elemental/issues/1124

## Pull requests
- [ ] https://github.com/silverstripe/silverstripe-linkfield/pull/90

sabina-talipova:
PR is merged. Issue was fixed.

PR #90, opened 2023-09-26
ENH Add versioning to Link

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-linkfield/issues/7


_config.php
michalkleiner, line 1: Does this file need any content at all?

src/Models/Link.php
```diff
@@ -32,6 +33,10 @@ class Link extends DataObject implements JsonData, Type
         'Title' => 'Varchar',
         'OpenInNew' => 'Boolean',
     ];
+    
+    private static array $extensions = [
+        Versioned::class
```
michalkleiner:
```suggestion
        Versioned::class,
```

_config.php
emteknetnz: I made this change just to pass phpcs. There's a bunch of other PRs linked in the parent issue that has the same change.

I'd delete it, but I'm doing lots of little PRs with a single focus to keep it clear what's changing rather than one big PR. The file will probably get deleted in a final tidy up PR

src/Models/Link.php
emteknetnz: Updated

tests/php/Models/LinkTest.php
```diff
@@ -37,13 +37,8 @@ protected function setUp(): void
 
         /** @var Image $image */
         $image = $this->objFromFixture(Image::class, 'image-1');
-        $image->setFromLocalFile(Director::baseFolder() . '/tests/resources/600x400.png');
+        $image->setFromLocalFile(dirname(dirname(dirname(__FILE__))) . '/resources/600x400.png');
```
emteknetnz, line 40: This doesn't affect versioning though it's required to run the unit test from the vendor folder

tests/php/Models/LinkTest.php
```diff
@@ -37,13 +37,8 @@ protected function setUp(): void
...
         $image->write();
-        $image->publishSingle();
-
-        /** @var SiteTree $page */
-        $page = $this->objFromFixture(SiteTree::class, 'page-1');
-        $page->publishSingle();
```
emteknetnz, line 46: I don't know why the previous unit test was publishing things and setting the stage to LIVE, it's not necessary for the purpose of what's being tested

tests/php/Models/LinkTest.php
```diff
@@ -250,14 +271,9 @@ public function linkTypeEnabledProvider(): array
      */
     public function testGetUrl(string $identifier, string $class, string $expected): void
     {
-        Versioned::withVersionedMode(function () use ($identifier, $class, $expected): void {
-            Versioned::set_stage(Versioned::LIVE);
```
emteknetnz, line 254: Versioning doesn't have any impact on this test (it's testing getURL()), and setting things to LIVE means we'd need to publish the Links before making assertions

emteknetnz:
Design feedback:
- Keep icons as they are
- Add versioned badges to the right of the descriptions, similar to what the site tree has
- `<span class="badge status-modified" title="Page has unpublished changes">Modified</span>`  `#cf3f00` `#fff2ea`
- Truncate title in PHP to a reasonable character limit (30?) and add '...' if needed
- Raise a card to implement text badges in elemental

emteknetnz:
Design review has implemented and approved

README.md
```diff
@@ -40,7 +40,22 @@ class Page extends SiteTree
...
-        'HasManyLinks' => Link::class
+        'HasManyLinks' => Link::class,
+    ];
+
+    private static array $owns = [
+        'HasOneLink',
+        'HasManyLinks',
+    ];
+
+    private static array $cascade_deletess = [
```
sabina-talipova:
```suggestion
    private static array $cascade_deletes = [
```

client/src/components/LinkPicker/LinkPickerTitle.js
```diff
@@ -12,11 +13,41 @@ const stopPropagation = (fn) => (e) => {
...
+
+const LinkPickerTitle = ({ id, title, description, versionState, typeTitle, onClear, onClick }) => {
+  const classes = {
+    'link-picker__link': true,
+    'form-control': true,
+  };
+  if (versionState) {
+    classes[` link-picker__link--${versionState}`] = true;
+  }
+  if (title.length > 30) {
```
sabina-talipova:
```suggestion
  if (title && title.length > 30) {
```

client/src/components/LinkPicker/LinkPickerTitle.js
```diff
@@ -12,11 +13,41 @@ const stopPropagation = (fn) => (e) => {
...
+    title = title.substring(0, 30) + '...';
```
sabina-talipova: Probably, we should decrease it up to 20 signs. "Modified" goes to the another string with long title.
Screenshot 2023-12-11 at 11 53 54 AM

sabina-talipova (changes requested):
Just tiny changes.

README.md
emteknetnz: Updated

client/src/components/LinkPicker/LinkPickerTitle.js
emteknetnz: During the design review we agreed on 30 so would prefer to keep it closer to that.

In split mode we need to go all the way down to 17 chars to get it to not wrap, which isn't enough chars IMO.

image

The bigger issue IMO that it goes over-bounds in split mode with 30 chars
image

Reducing this to 25 resolves this
image

I'll double check with design team if they're prefer 25 with wrap or 17 without wrap

emteknetnz: Chatted with design team, agreed on 25 with wrap though ensure that badge isn't indented if it wraps

image

client/src/components/LinkPicker/LinkPickerTitle.js

client/src/components/LinkPicker/LinkPickerTitle.js
michalkleiner: Put 25 `M` characters and it will do the same.. Does it need to be character limit? CSS can put the ellipsis there automatically based on width of the container.

sabina-talipova: I think we can extract this issue in another ticket, since there is another similar issue with long external link. Probably, we should review link field styles.

sabina-talipova: Open new issue for UX issue
- https://github.com/silverstripe/silverstripe-linkfield/issues/138

sabina-talipova (approved):
LGTM. Test locally. Works perfect.

Issue #12, opened 2020-11-23
Permisson model


maxime-rainville:
# Story
As a **Site Owner** I want a **proper access control to be enforced on Link data** so that **I have confidence that only autorised user can view and edit link**.

## Acceptance criteria
- Sensible CanView/CanDelete/CanEdit/CanCreate method have been created.
- Can* method are driven by their owner/parent. (e.g. If you can view the page that owns the link, you can view the link as well)
- If at all possible, make sure `Link` is aware of the owner trying to create it before calling `canCreate()` on it, so we can correctly rely on the parent's `canEdit()` permissions for creation
- Link permission can be customised with Extensions
- Permissions have unit tests
 
## Notes
- There's some UX questions about what would happen if a link points to a DataObject the current user is now allowed to view. Those concerns will be dealth with in a separate card.

## POC PRs
- https://github.com/silverstripe/silverstripe-linkfield/pull/101
- https://github.com/silverstripe/silverstripe-linkfield/pull/102

## PRs
- https://github.com/silverstripe/silverstripe-linkfield/pull/134

sabina-talipova:
PR was merged. Close.

Issue #5, opened 2020-11-23
Reverse ownership


maxime-rainville:
## Story
As a **developer** I want **Link objects to have a sensible ownership model** so that **Link objects behave in a sensible way when versioned operation are performed**.

## Acceptance criteria
- We have validated what the ownership relations between Link and their owner object should be.
- Usage doc provides clear guidance on how to configure ownership when creating Link relations.
- Ownership model can accommodate situation where link is in Has_many or a Has_One relation.
- The scenario where two "HAS_Many" link relations are managed on the same DataObject is considered.

## Note
The links belongs to a single owner relationship.

### Option one:

- ~https://github.com/silverstripe/silverstripe-linkfield/pull/101~ rejected, use option 2

### Option two:
NOTE: This option could be mostly implemented in framework if we have some new way to define it (e.g. `has_one_special`) - that's necessary so we know when we should have and use the new dbfield which existing has_one relations won't have. 
- https://github.com/silverstripe/silverstripe-framework/pull/10973
- https://github.com/silverstripe/silverstripe-linkfield/pull/102

## INSTALLER CI RUN
https://github.com/creative-commoners/silverstripe-installer/actions/runs/7160809301
Note the `SilverStripe\CMS\Tests\Model\SiteTreeTest::testRelativeLink` failure is pre-existing. See https://github.com/silverstripe/.github/issues/169

## PRs
- https://github.com/silverstripe/silverstripe-framework/pull/11084
- https://github.com/silverstripe/developer-docs/pull/417
- https://github.com/silverstripe/silverstripe-versioned/pull/427
- https://github.com/silverstripe/silverstripe-linkfield/pull/127

GuySartorelli:
Added an alternative solution.

michalkleiner:
No input as to the issue itself, just o note on the story language — 'sensible' can be quite ambiguous and it's not clear (to me) how to quantify that and what it should express/mean.

maxime-rainville:
I had a glance at Guy's solution. I didn't think it was obviously wrong or horrible ... although it seems like a lot of effort to accommodate a rather simple use case.

In the spirit of iteration, what would happen if we just declared that only one link list can be tied to one record? We could wait to see if anyone actually has a use case for having two or more link lists. Could support for multiple Link List on one DataObject retroactively added later on?

GuySartorelli:
> We could wait to see if anyone actually has a use case for having two or more link lists.

[...]

PR #101, opened 2023-10-09
NEW Linkfield ownership

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-linkfield/issues/5

Also implements https://github.com/silverstripe/silverstripe-linkfield/issues/12

**In peer review because looking for initial feedback - have I overlooked something really obvious?**

This is a draft PR that's just testing out the idea of using a LinkArea to manage the has_many relationships similar to ElementalArea in elemental. Advantage of using the extra layer is that it allows you to effectively add multiple "has_many" relations of Link::class and have them correctly assigned to each of the has_many relations.

Documentation would instruct you to add you Links and LinkArea's to $has_one + $owns + $cascade_deletes + $cascase_duplicates

A Page in a project would look something like this if it had two Link's and two LinkArea's

```php
class Page extends SiteTree
{
        'MyLinkOne' => Link::class,
        'MyLinkTwo' => Link::class,
        'MyLinkAreaOne' => LinkArea::class,
        'MyLinkAreaTwo' => LinkArea::class,
    ];

    private static array $owns = [
        'MyLinkOne',
        'MyLinkTwo',
        'MyLinkAreaOne',
        'MyLinkAreaTwo',
    ];

    private static array $cascade_deletes = [
        'MyLinkOne',
        'MyLinkTwo',
        'MyLinkAreaOne',
        'MyLinkAreaTwo',
    ];

    private static array $cascade_duplicates = [
        'MyLinkOne',
        'MyLinkTwo',
        'MyLinkAreaOne',
        'MyLinkAreaTwo',
    ];
}

```

Adding multiple has_many's of the same type normally isn't possible because you'll have something like this:

```
Page
  $has_many = [
    MyThings => MyDataObject::class,
    MoreThings => MyDataObject::class,
  ]

MyDataObject
  $has_one = [
    Page => Page::class
  ];
```

And you'll get $myDataObject->PageID = 123 - but there's no way to know if it should belong to the 'MyThings' or the 'MoreThings' relationship on Page

The official way to solve "multiple has_manys of the same type" this is with [dot notation](https://docs.silverstripe.org/en/5/developer_guides/model/relations/#dot-notation) where the the corresponding has_one relationship is suffixed to the classname of has_many relationship. In this instance it would look something like this (excuse my use of RelOne and RelTwo, I couldn't think of proper naming here):

```
Page
  $has_many = [
    MyThings => MyDataObject::class . '.RelOne',
    MoreThings => MyDataObject::class . '.RelTwo',
  ]

MyDataObject
  $has_one = [
    RelOne => Page::class,
    RelTwo => Page::class,
  ];
```

However in the linkfield context we cannot use dot-notation, because "MyDataObject" would be "Link" and we can't set the $has_one's on Link since it has no knowledge of project code. We'd need to ask people to add the $has_ones to Link via extensions which is very clunky.

Note there's somethings to do with unit testing that needs to be extracted from https://github.com/silverstripe/silverstripe-linkfield/pull/90 before closing that PR

### DIFF
--- a/_config/config.yml
+++ b/_config/config.yml
@@ -21,3 +21,7 @@ SilverStripe\CMS\Forms\AnchorSelectorField:
 SilverStripe\LinkField\Form\FormFactory:
   extensions:
     - SilverStripe\LinkField\Extensions\FormFactoryExtension
+
+SilverStripe\ORM\DataObject:
+  extensions:
+    - SilverStripe\LinkField\Extensions\DataObjectExtension
--- a/src/Extensions/DataObjectExtension.php
+++ b/src/Extensions/DataObjectExtension.php
@@ -1,0 +1,47 @@
+<?php
+
+namespace SilverStripe\LinkField\Extensions;
+
+use SilverStripe\ORM\DataExtension;
+use SilverStripe\LinkField\Models\Link;
+use SilverStripe\LinkField\Models\LinkArea;
+
+class DataObjectExtension extends DataExtension
+{
+    public function onAfterWrite(): void
+    {
+        // Using onAfterWrite instead of onBeforeWrite to ensure that $this->owner->ID is not zero when creating new objects
+        parent::onAfterWrite();
+        foreach ($this->owner->hasOne() as $relation => $relationClassName) {
+            $relationField = $relation . 'ID';
+            $relationID = $this->owner->$relationField;
+            if (!$relationID) {
+                continue;
+            }
+            if (!is_a($relationClassName, Link::class, true) && !is_a($relationClassName, LinkArea::class, true)) {
+                continue;
+            }
+            // skip for the has_one:LinkArea relation on Link
+            if (is_a($this->owner, Link::class) && $relation === 'LinkArea') {
+                continue;
+            }
+            $relationObj = $relationClassName::get()->byID($relationID);
+            if ($relationObj === null) {
+                // could throw an Exception here, though not sure how if user would be able to fix it
+                continue;
+            }
+            $doWrite = false;
+            if ($relationObj->OwnerID !== $this->owner->ID) {
+                $relationObj->OwnerID = $this->owner->ID;
+                $doWrite = true;
+            }
+            if ($relationObj->OwnerClassName !== $this->owner->ClassName) {
+                $relationObj->OwnerClassName = $this->owner->ClassName;
+                $doWrite = true;
+            }
+            if ($doWrite) {
+                $relationObj->write();
+            }
+        }
+    }
+}
--- a/src/Extensions/LinkObjectExtension.php
+++ b/src/Extensions/LinkObjectExtension.php
@@ -1,0 +1,76 @@
+<?php
+
+namespace SilverStripe\LinkField\Extensions;
+
+use SilverStripe\LinkField\Models\Link;
+use SilverStripe\ORM\DataExtension;
+use SilverStripe\ORM\DataObject;
+
+/**
+ * This is only intended to be added to Link and LinkArea
+ * Implemented as an extension rather than base class so it doesn't create an extra base table that needs to be joined
+ */
+class LinkObjectExtension extends DataExtension
+{
+    private static array $db = [
+        'OwnerID' => 'Int',
+        'OwnerClassName' => 'Varchar',
+    ];
+
+    public function canView($member = null)
+    {
+        return $this->canCheck('canView', $member);
+    }
+
+    public function canCreate($member = null)
+    {
+        return $this->canCheck('canCreate', $member);
+    }
+
+    public function canEdit($member = null)
+    {
+        return $this->canCheck('canEdit', $member);
+    }
+
+    public function canDelete($member = null)
+    {
+        return $this->canCheck('canDelete', $member);
+    }
+
+    private function canCheck(string $canMethod, $member)
+    {
+        if (!$member) {
+            $member = Security::getCurrentUser();
+        }
+        $extended = $this->extendedCan($canMethod, $member);
+        if ($extended !== null) {
+            return $extended;
+        }
+        $owner = $this->getOwningDataObject();
+        if ($owner) {
+            return $owner->$canMethod($member);
+        }
+        return parent::$canMethod($member);
+    }
+
+    private function getOwningDataObject(): ?DataObject
+    {
+        // Thismethod is not called getOwner() because of existing Extension::getOwner() method
+        //
+        // If this is a Link, and LinkArea is set on it return that
+        if (is_a($this->owner, Link::class, true)) {
+            $linkArea = $this->owner->LinkArea();
+            if ($linkArea && $linkArea->exists()) {
+                return $linkArea;
+            }
+        }
+        // Otherwise look for the ownerID + ownerClassName
+        // These are set in DataObjectExtension::onAfterWrite()
+        $ownerID = $this->owner->OwnerID;
+        $ownerClassName = $this->owner->OwnerClassName;
+        if ($ownerID === 0 || $ownerClassName === '') {
+            return null;
+        }
+        return $ownerClassName::get()->byID($ownerID);
+    }
+}
--- a/src/Models/Link.php
+++ b/src/Models/Link.php
@@ -16,6 +16,9 @@ use SilverStripe\LinkField\Type\Type;
 use SilverStripe\ORM\DataObject;
 use SilverStripe\ORM\FieldType\DBHTMLText;
 use SilverStripe\View\Requirements;
+use SilverStripe\LinkField\Extensions\LinkObjectExtension;
+use SilverStripe\Versioned\Versioned;
+use SilverStripe\LinkField\Models\LinkArea;
 
 /**
  * A Link Data Object. This class should be a subclass, and you should never directly interact with a plain Link
@@ -31,6 +34,15 @@ class Link extends DataObject implements JsonData, Type
     private static array $db = [
         'Title' => 'Varchar',
         'OpenInNew' => 'Boolean',
+    ];
+
+    private static array $has_one = [
+        'LinkArea' => LinkArea::class
+    ];
+
+    private static array $extensions = [
+        Versioned::class,
+        LinkObjectExtension::class,
     ];
 
     /**
--- a/src/Models/LinkArea.php
+++ b/src/Models/LinkArea.php
@@ -1,0 +1,34 @@
+<?php
+
+namespace SilverStripe\LinkField\Models;
+
+use SilverStripe\LinkField\Models\Link;
+use SilverStripe\LinkField\Extensions\LinkObjectExtension;
+use SilverStripe\ORM\DataObject;
+use SilverStripe\Versioned\Versioned;
+
+class LinkArea extends DataObject
+{
+    private static $table_name = 'LinkField_LinkArea';
+
+    private static array $has_many = [
+        'Links' => Link::class,
+    ];
+
+    private static array $owns = [
+        'Links',
+    ];
+
+    private static array $cascade_deletes = [
+        'Links',
+    ];
+
+    private static array $cascade_duplicates = [
+        'Links',
+    ];
+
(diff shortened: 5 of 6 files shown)


emteknetnz:
Just had an offline chat with Guy - we could get rid of DataObject extension and just put a `$has_one => [Owner]` on Link/LinkArea if we instead made the Page / project level a `$belongs_to` instead of a `$has_one`.  However this then becomes a bit of a usability issue because website developers are very much used to using $has_one, and $belongs_to is the weird one that people often don't bother to implement.  This means we'd be turning LinkField's API into something a bit non-standard

_config/config.yml
```diff
@@ -21,3 +21,7 @@ SilverStripe\CMS\Forms\AnchorSelectorField:
 SilverStripe\LinkField\Form\FormFactory:
   extensions:
     - SilverStripe\LinkField\Extensions\FormFactoryExtension
+
+SilverStripe\ORM\DataObject:
+  extensions:
+    - SilverStripe\LinkField\Extensions\DataObjectExtension
```
ChloeSartorelli, line 27: Let people apply it themselves to any classes where they want to use links (the same way elemental doesn't apply its extension to `DataObject` or to `SiteTree` by default).

src/Extensions/LinkObjectExtension.php
```diff
@@ -0,0 +1,76 @@
...
+/**
+ * This is only intended to be added to Link and LinkArea
+ * Implemented as an extension rather than base class so it doesn't create an extra base table that needs to be joined
+ */
+class LinkObjectExtension extends DataExtension
+{
+    private static array $db = [
+        'OwnerID' => 'Int',
+        'OwnerClassName' => 'Varchar',
+    ];
```
ChloeSartorelli, line 18:
```suggestion
    private static array $jas_one = [
        'Owner' => DataObject::class,
    ];
```
This creates the same db fields (but with the correct types, dbforeignkey and dbclassname) and gives us the ability to go `$this->Owner()`

This should be declared directly on `Link` and `LinkArea` though. No need for it here.

ChloeSartorelli: Also should probably be `Parent` instead of `Owner`

src/Extensions/LinkObjectExtension.php
```diff
@@ -0,0 +1,76 @@
...
+
+use SilverStripe\LinkField\Models\Link;
+use SilverStripe\ORM\DataExtension;
+use SilverStripe\ORM\DataObject;
+
+/**
+ * This is only intended to be added to Link and LinkArea
+ * Implemented as an extension rather than base class so it doesn't create an extra base table that needs to be joined
+ */
+class LinkObjectExtension extends DataExtension
```
ChloeSartorelli, line 13: Don't make this an extension. If you want to share the can* methods just use a trait which is more performant. There's nothing here that needs to be in an extension.

src/Extensions/LinkObjectExtension.php
```diff
@@ -0,0 +1,76 @@
...
+        }
+        // Otherwise look for the ownerID + ownerClassName
+        // These are set in DataObjectExtension::onAfterWrite()
+        $ownerID = $this->owner->OwnerID;
+        $ownerClassName = $this->owner->OwnerClassName;
+        if ($ownerID === 0 || $ownerClassName === '') {
+            return null;
+        }
+        return $ownerClassName::get()->byID($ownerID);
+    }
```
ChloeSartorelli, line 75: This won't be needed on `LinkArea` at all because of the has_one.

On `Link`, the `Owner` relation will be the `LinkArea`, so this method will still be useful there. But it should be like this:

```php
public function Owner()
{
    $owner = $this->getComponent('Owner');

    // If $owner is the LinkArea, return the underlying record that owns the area.
    if ($owner instanceof LinkArea) {
        $owner = $owner->Owner();
    }
    return $owner;
}
```

Unfortunately we can't call this `getOwner()` because the magic that turns relations into methods doesn't check for getters.

With this method, we can rely on `$this->Owner()` in `Link` always being whatever ultimately owns the link - the `LinkArea` record is completely transparent.

src/Extensions/LinkObjectExtension.php
```diff
@@ -0,0 +1,76 @@
...
+    private function canCheck(string $canMethod, $member)
+    {
+        if (!$member) {
+            $member = Security::getCurrentUser();
+        }
+        $extended = $this->extendedCan($canMethod, $member);
+        if ($extended !== null) {
+            return $extended;
+        }
+        $owner = $this->getOwningDataObject();
```
ChloeSartorelli, line 49:
```suggestion
        $owner = $this->Owner();
```

src/Models/Link.php
```diff
@@ -33,6 +36,15 @@ class Link extends DataObject implements JsonData, Type
         'OpenInNew' => 'Boolean',
     ];
 
+    private static array $has_one = [
+        'LinkArea' => LinkArea::class
+    ];
```
ChloeSartorelli, line 41:
```suggestion
    private static array $has_one = [
        'Owner' => DataObject::class
    ];
```
As per above comments. Again, probably this should be `Parent`, but I'll leave everything as `Owner` for this pass.

src/Models/Link.php
```diff
@@ -33,6 +36,15 @@ class Link extends DataObject implements JsonData, Type
...
+
+    private static array $extensions = [
+        Versioned::class,
+        LinkObjectExtension::class,
```
ChloeSartorelli, line 45:
```suggestion
```
Either add all the can* methods in here or apply them as a trait.

src/Models/Link.php
```diff
@@ -33,6 +36,15 @@ class Link extends DataObject implements JsonData, Type
         'OpenInNew' => 'Boolean',
     ];
 
+    private static array $has_one = [
+        'LinkArea' => LinkArea::class
+    ];
+
+    private static array $extensions = [
+        Versioned::class,
```
ChloeSartorelli, line 44: Why versioned?

src/Models/LinkArea.php
```diff
@@ -0,0 +1,34 @@
...
+        'Links',
+    ];
+
+    private static array $cascade_duplicates = [
+        'Links',
+    ];
+
+    private static array $extensions = [
+        Versioned::class,
+        LinkObjectExtension::class,
```
ChloeSartorelli, line 32:
```suggestion
```
Either add all the can* methods in here or apply them as a trait.

src/Models/LinkArea.php
```diff
@@ -0,0 +1,34 @@
...
+    private static array $cascade_deletes = [
+        'Links',
+    ];
+
+    private static array $cascade_duplicates = [
+        'Links',
+    ];
+
+    private static array $extensions = [
+        Versioned::class,
```
ChloeSartorelli, line 31: Why versioned?

src/Extensions/DataObjectExtension.php
```diff
@@ -0,0 +1,47 @@
...
+            if (!$relationID) {
+                continue;
+            }
+            if (!is_a($relationClassName, Link::class, true) && !is_a($relationClassName, LinkArea::class, true)) {
+                continue;
+            }
+            // skip for the has_one:LinkArea relation on Link
+            if (is_a($this->owner, Link::class) && $relation === 'LinkArea') {
+                continue;
+            }
```
ChloeSartorelli, line 27:
```suggestion
```
This extension shouldn't be applied directly to `Link` or to `DataObject`, so we don't need this - unless we expect a link to ever have a child link? But I'd say we don't want to support nesting links.

src/Extensions/DataObjectExtension.php
```diff
@@ -0,0 +1,47 @@
...
+            $doWrite = false;
+            if ($relationObj->OwnerID !== $this->owner->ID) {
+                $relationObj->OwnerID = $this->owner->ID;
+                $doWrite = true;
+            }
+            if ($relationObj->OwnerClassName !== $this->owner->ClassName) {
+                $relationObj->OwnerClassName = $this->owner->ClassName;
+                $doWrite = true;
+            }
```
ChloeSartorelli, line 41: I don't like that we're essentially storing the has_one in two places in the DB here. We're really working around a fundamental limitation in framework.
I would prefer that we either:
1. Explicitly document that people _have_ to use `belongs_to` for their links and link areas, OR
2. Implement a way in framework to say "these has_one relations are _the same_ relation, so only store it once!" effectively treating one of them as a belongs_to, OR
3. Make framework store has_many in its own table like many_many, which would allow using a single belongs_to in `Link` to an arbitrary number of has_many or has_one relations on other classes (which means we wouldn't need `LinkArea` either) (would be a new major of framework, or a new relation type, or a new behaviour that only happens when framework identifies that the has_many is trying to link up to a belongs_to))

All of the above do away with needing this extension altogether.

GuySartorelli (commented):
I have reviewed this assuming that the approach is fundamentally solid.

Another way we could approach this is to reconsider how has_many relations work.
- We could add a new `OwnerRelation` DB field. This would go hand-in hand with the `OwnerID` and `OwnerClass` fields that will already exist from the polymorphic has_one.
- When the link field saves the link, it saves the _name_ of the relation that owns it. This way we have a way to identify which has_many or has_one was used to save the given record.
- We could implement this somewhere in framework, and then in `DataObject::getComponents()` if this special new relation type is being used, add a `->filter("{$hasOneName}Relation", $hasManyName)` so it only fetches the records for that relation.
- If we don't implement any of this in framework, we can add an extension hook to `DataObject::getComponents()` (there's already one in `getManyManyComponents()` so it makes sense to add one here too) and add that filter via the extension hook.

There's probably a bunch of stuff about that approach that seems obvious to me that isn't obvious to anyone else so please do ask clarification questions.

src/Extensions/LinkObjectExtension.php
emteknetnz: > This won't be needed on LinkArea at all because of the has_one.

I'm not sure? The has_one is on an artibary DataObject (usually Page), so the LinkArea needs this as a way to find it's Owner/Parent?

> On Link, the Owner relation will be the LinkArea

Not if it's not using a LinkArea, and instead it's a has_one on an arbitary DataObject (usually Page).

General idea for this PR was that it allows Links to be used as has_one's on Page's, or as has_many's via LinkArea's

src/Models/Link.php
emteknetnz: Linked [issue](https://github.com/silverstripe/silverstripe-linkfield/issues/5) says "versioned operations" - you need to make Link Versioned to do versioned operations :-)

src/Models/LinkArea.php
emteknetnz: Replied above on Link

emteknetnz:
Have chatted offline with Guy re his ideas around adding a "third has-many column" that is essentially a replacement for dot-notation. If that works it would remove the need for the intermediary LinkArea table, which would be great performance wise.  So he'll explore that further and if that works we'll ditch this approach

src/Extensions/LinkObjectExtension.php
emteknetnz: The private static $db cannot be used as a trait, because it'll clash with the private static $db on Link.  You'll get this error:

PHP Fatal error:  Link and LinkObjectTrait define the same property ($db) in the composition of Link. However, the definition differs and is considered incompatible.

The Extension class is basically a heavier weight "Silverstripe trait" and will nicely merge private statics

ChloeSartorelli: > The private static $db cannot be used as a trait, because it'll clash with the private static $db on Link.

Correct. As suggested in https://github.com/silverstripe/silverstripe-linkfield/pull/101#discussion_r1351370521 the `has_one` should be declared on the distinct classes instead. Ideally, so will the can* methods but I understand not wanting to repeat them

emteknetnz:
Closing in favor of https://github.com/silverstripe/silverstripe-linkfield/pull/102

Issue #105, opened 2023-10-19
Refactor LinkField around FormSchema


maxime-rainville:
As a CMS Squad, we want the LinkField to use best practices to manage its communication with the backend so it's easy to maintain in the long run.

## Acceptance criteria
- Only the Link ID is communicated to the front end (rather than a big JSON blob)
- Initial Link Title and Description is fetched via a scaffolded GraphQL schema. 
- The form schema in the Link Modal is retrieved via the ID of the link being edited.
- When the Link modal is submitted, the link is created/updated on the spot.
- LinkField is updated to remove the logic to handle scenarios where the type of a link is changed.

## Exclude
- Form schema validation

## PRs
- https://github.com/silverstripe/silverstripe-linkfield/pull/108

maxime-rainville:
We initially wanted to block this card on "Standardised how we create API/FormSchema endpoint in the CMS". We decided to reverse the block and use this card as a POC. We'll learn lessons from this card and reimplement it later on to use the new standard.

maxime-rainville:
Achievement unlocked!

Issue #15, opened 2020-11-23
Controlling link type per field


maxime-rainville:
# Story
As a **Developer** I want **the ability to control what Link Type are accepted on a given Link Field** so that **I can prevent user from selecting nonsensical link type**.

## Acceptance criteria
- Individual LinkField have a configuration option to exclude/include specific link type.
- Link type are auto detected from the child class of Link.
- "Link Registry" concept is removed.
 
## Note
- A good approch to implementing this is that possible link types are passed as props on the React component rather than an GraphQL query. AnyFiled has a good example of this implementation.
- The current implementation passes link type via a GraphQL query that can not be customised for individual links.
- There's a implementation of the allow/deny link type on the [gorriecoe/silverstripe-linkfield](https://github.com/elliot-sawyer/silverstripe-linkfield) module.
 
## Things that were considered and rejected
- LinkField that point to a relation with a specific Link Type only accept this Type
- Allow specific link type to be turn off by default

## Pull requests
- https://github.com/silverstripe/silverstripe-linkfield/pull/86 (Experimental / Draft)
- https://github.com/silverstripe/silverstripe-linkfield/pull/137

maxime-rainville:
I think removing registry through this card makes a lot of sense because managing possible link type is intertwined with controlling what type of links can be created via a specific field.

AnyField does the "Possible link types are passed as props on the React component" pretty well I think. So we should look there for inspiration.

I'm not sure about those two ACs.
- (LinkField that point to a relation with a specific Link Type only accept this Type)
  - That's very intuitive for the developer, but getting AnyField to autodetech relationship type was a massive pain on the back side.
- (Allow specific link type to be turn off by default)
  - The idea here is that maybe you have a very esoteric LinkType that most LinkField shouldn't accept.

maxime-rainville:
All done.

PR #108, opened 2023-10-24
NEW LinkFieldController to handle FormSchema

emteknetnz:
You can use the following to quickly create a page with a link field on it

**Page.php**
```php
<?php

use SilverStripe\CMS\Model\SiteTree;
use SilverStripe\LinkField\Form\LinkField;
use SilverStripe\LinkField\Models\Link;

class Page extends SiteTree
{
    private static $db = [];

    private static $has_one = [
        'MyLink' => Link::class,
    ];

    {
        $fields = parent::getCMSFields();
        $fields->insertAfter('Content', new LinkField('MyLink'));
        return $fields;
    }
}
```

You can test form validation of PhoneLink using invalid values of "x" and "y" with the following extension:

**extensions.yml**
```yml
SilverStripe\LinkField\Models\PhoneLink:
  extensions:
    - PhoneLinkExtension

```

**PhoneLinkExtension.php**
```php
<?php

use SilverStripe\ORM\DataExtension;
use SilverStripe\ORM\ValidationResult;
use SilverStripe\Forms\CompositeValidator;
use SilverStripe\Forms\Validator;

class PhoneLinkExtension extends DataExtension
{
    public function validate(ValidationResult $validationResult)
    {
        $phone = $this->owner->Phone;
        if ($phone == 'x') {
            $validationResult->addFieldError('Phone', 'Cannot be x - DataObject::validate()');
        }
        return $validationResult;
    }

    public function updateCMSCompositeValidator(CompositeValidator $compositeValidator): void
    {
        $compositeValidator->addValidator(new class extends Validator {
            public function php($data): bool
            {
                $valid = true;
                $phone = $data['Phone'];
                if ($phone == 'y') {
                    $valid = false;
                    $this->validationError('Phone', 'Cannot be y -- DataObject::getCMSCompositeValidator()');
                }
                return $valid;
            }
        });
    }
}
```

## Parent issue
- #105


src/Controllers/LinkFieldController.php
```diff
@@ -0,0 +1,298 @@
...
+        $response->setBody(json_encode(['success' => true]));
+        return $response;
+    }
+
+    /**
+     * I tried to use GraphQL but it stopped making requests once I switched it to
+     * passing an ID instead of the entire JSON blob of the link - seems like there's
+     * some logic where passing the same param means don't refetch?
+     */
+    public function linkData(): HTTPResponse
```
emteknetnz, line 252: I tried to use GraphQL but it stopped making requests once I switched it to only using ID instead of the entire JSON blob of the link to get the description (have since updated endpoint to return all data, not just description) - seems like there's logic (in apollo?) where passing the same param means don't refetch?

src/Controllers/LinkFieldController.php
```diff
@@ -0,0 +1,298 @@
...
+        $response->setStatusCode(200);
+        return $response;
+    }
+
+    /**
+     * I tried to use GraphQL but could work out apollo so make this instead
+     * this query did work in graphql ide
+     * mutation { deleteLinks(ids: [44]) }
+     */
+    public function linkDelete(): HTTPResponse
```
emteknetnz, line 233: I tried to use GraphQL but could work out apollo so just made this little endpoint instead.  The following query did work in graphql ide - `mutation { deleteLinks(ids: [44]) }`

tests/php/Models/LinkTest.php
```diff
@@ -37,7 +36,7 @@ protected function setUp(): void
 
         /** @var Image $image */
         $image = $this->objFromFixture(Image::class, 'image-1');
-        $image->setFromLocalFile(Director::baseFolder() . '/tests/resources/600x400.png');
+        $image->setFromLocalFile(dirname(dirname(dirname(__FILE__))) . '/resources/600x400.png');
```
emteknetnz, line 39: This is required to get the unit-tests to run from both the vendor folder and non-vendor folder installs

src/Controllers/LinkFieldController.php
```diff
@@ -0,0 +1,295 @@
...
+            $this->jsonError(403, 'Unauthorized');
+        }
+        return $this->createLinkForm($link);
+    }
+
+    /**
+     * Get data for a Link
+     * /admin/linkfield/data/<LinkID>
+     */
+    public function linkData(): HTTPResponse
```
emteknetnz, line 84: I tried to use GraphQL but it stopped making requests once I switched it to only using ID instead of the entire JSON blob of the link to get the description (have since updated endpoint to return all data, not just description) - seems like there's logic (in apollo?) where passing the same param means don't refetch?

src/Controllers/LinkFieldController.php
```diff
@@ -0,0 +1,295 @@
...
+        $response = $this->getResponse();
+        $response->addHeader('Content-type', 'application/json');
+        $data = $link->jsonSerialize();
+        $data['description'] = $link->getDescription();
+        $response->setBody(json_encode($data));
+        return $response;
+    }
+
+    /**
+     * Delete a Link
```
emteknetnz, line 103: I tried to use GraphQL but I couldn't work out how to use apollo so just made this little endpoint instead. The following query did work in graphql ide - mutation { deleteLinks(ids: [44]) }

src/Controllers/LinkFieldController.php
```diff
@@ -0,0 +1,295 @@
...
+            }
+        }
+        if (!$link->canView()) {
+            $this->jsonError(403, 'Unauthorized');
+        }
+        return $this->createLinkForm($link);
+    }
+
+    /**
+     * Get data for a Link
```
emteknetnz, line 85: I tried to use GraphQL but it stopped making requests once I switched it to only using ID instead of the entire JSON blob of the link to get the description (have since updated endpoint to return all data, not just description) - seems like there's logic (in apollo?) where passing the same param means don't refetch?

_config/config.yml
```diff
@@ -1,23 +1,6 @@
...
-  extensions:
-    - SilverStripe\LinkField\Extensions\ModalController
-
-SilverStripe\Forms\TreeDropdownField:
-  extensions:
-    - SilverStripe\LinkField\Extensions\AjaxField
-
-SilverStripe\CMS\Forms\AnchorSelectorField:
-  extensions:
-    - SilverStripe\LinkField\Extensions\AjaxField
```
maxime-rainville, line 19: Note to self: make sure that TreeDropDown and AnchorField still works.

_graphql/queries.yml
```diff
@@ -1,6 +1,3 @@
-'readLinkDescription(dataStr: String!)':
-  type: LinkDescription
-  resolver: ['SilverStripe\LinkField\GraphQL\LinkDescriptionResolver', 'resolve']
 'readLinkTypes(keys: [ID])':
```
maxime-rainville, line 1: Note to self: Make sure there's a card to ditch the registry and get the link types to be communicated as props.

client/src/components/LinkField/LinkField.js
```diff
@@ -1,90 +1,141 @@
...
+/**
+ * value - initial value of the field passed from JsonField
+ * onChange - callback function passed from JsonField
+ * types - injected by the GraphQL query
+ * actions - object of redux actions
+ */
+const LinkField = ({ value, onChange, types, actions }) => {
+  const [typeKey, setTypeKey] = useState(value.typeKey || '');
+  const [linkID, setLinkID] = useState(value.ID || 0);
+  const [data, setData] = useState(value);
```
maxime-rainville: Setting the initial value of `useState` hook from a prop is wrong.

`onChange` should be updating the parent state and the information should flow back down there.

client/src/components/LinkField/LinkField.js
```diff
@@ -1,90 +1,141 @@
...
+  const [editing, setEditing] = useState(false);
 
-const LinkField = ({ id, loading, Loading, data, LinkPicker, onChange, types, linkDescription, ...props }) => {
-  if (loading) {
-    return <Loading />;
-  }
+  /**
+   * Call back used by LinkModal after the form has been submitted and the response has been received
+   */
+  const onModalSubmit = async (modalData, action, submitFn) => {
```
maxime-rainville, line 30: Need to handle other errors beyond validation. (e.g. Session time-out, not enough permission, random 500)

client/src/components/LinkField/LinkField.js
```diff
@@ -1,90 +1,141 @@
...
-  const [newTypeKey, setNewTypeKey] = useState('');
+      // update parent JsonField data id - this is required to update the underlying <input> form field
+      // so that the Page (or other parent DataObject) gets the Link relation ID set
+      onChange(id);
 
-  const onClear = (event) => {
-    if (typeof onChange !== 'function') {
-      return;
+      // success toast
+      actions.toasts.success('Saved link');
```
maxime-rainville, line 52: Localise this.

client/src/components/LinkField/LinkField.js
```diff
@@ -1,90 +1,141 @@
...
-  let title = data ? data.Title : '';
+  /**
+   * Call back used by LinkPicker when the 'Clear' button is clicked
+   */
+  const onClear = () => {
+    const endpoint = `${Config.getSection(section).form.linkForm.deleteUrl}/${linkID}`;
+    // CSRF token 'X-SecurityID' headers needs to be present for destructive requests
+    backend.delete(endpoint, {}, { 'X-SecurityID': Config.get('SecurityID') })
+      .then(() => {
+        actions.toasts.success('Deleted link');
```
maxime-rainville, line 66: Localise this.

client/src/components/LinkField/LinkField.js
```diff
@@ -1,90 +1,141 @@
...
+      })
+      .catch(() => {
+        actions.toasts.error('Failed to delete link');
```
maxime-rainville, line 69: Localise this.

client/src/components/LinkField/LinkField.js
```diff
@@ -1,90 +1,141 @@
...
+      });
+
+    // update component state
+    setTypeKey('');
+    setLinkID(0);
+    setData({});
```
maxime-rainville: This part shouldn't be needed. Only `onChange`.

client/src/components/LinkField/LinkField.js
```diff
@@ -1,90 +1,141 @@
...
+    linkID,
     data
   };
 
-  const handlerName = modalType ? modalType.handlerName : 'FormBuilderModal';
-  const LinkModal = loadComponent(`LinkModal.${handlerName}`);
+  // read data from endpoint and update component state
+  useEffect(() => {
+    if (!editing && linkID) {
+      const endpoint = `${Config.getSection(section).form.linkForm.dataUrl}/${linkID}`;
```
maxime-rainville, line 117: We are getting this URL to our section in multiple places. I'm thinking it might be worthwhile having it in one spot.

client/src/components/LinkPicker/LinkPicker.js
```diff
@@ -1,30 +1,33 @@
...
 
 LinkPicker.propTypes = {
   ...LinkPickerMenu.propTypes,
-  link: PropTypes.shape(LinkPickerTitle.propTypes),
+  title: PropTypes.string,
+  description: PropTypes.string,
+  typeTitle: PropTypes.string,
   onEdit: PropTypes.func,
   onClear: PropTypes.func,
+  onSelect: PropTypes.func
```
maxime-rainville: Seems to me like most of these things should be required.

client/src/components/LinkPicker/LinkPicker.js
```diff
@@ -1,30 +1,33 @@
...
 import LinkPickerTitle from './LinkPickerTitle';
-import LinkType from 'types/LinkType';
 
-const LinkPicker = ({ types, onSelect, link, onEdit, onClear }) => (
-  <div
-    className={classnames('link-picker', 'form-control', {'link-picker--selected': link})}>
-    {link === undefined && <LinkPickerMenu types={types} onSelect={onSelect} /> }
-    {link && <LinkPickerTitle {...link} onClear={onClear} onClick={() => link && onEdit && onEdit(link)}/>}
+const LinkPicker = ({ title, description, typeTitle, types, onSelect, onEdit, onClear }) => (
+  <div className={classnames('link-picker', 'form-control', {'link-picker--selected': typeTitle ? true : false})}>
```
maxime-rainville, line 9:
```suggestion
  <div className={classnames('link-picker', 'form-control', {'link-picker--selected': !!typeTitle})}>
```

client/src/components/LinkPicker/LinkPickerTitle.js
```diff
@@ -30,7 +29,7 @@ const LinkPickerTitle = ({ title, type, description, onClear, onClick }) => (
 
 LinkPickerTitle.propTypes = {
   title: PropTypes.string.isRequired,
-  type: LinkType,
+  linkTypeTitle: PropTypes.string,
   description: PropTypes.string,
   onClear: PropTypes.func,
   onClick: PropTypes.func
```
maxime-rainville: Most of these props should be required.

client/src/components/LinkPicker/LinkPickerTitle.js
```diff
@@ -30,7 +29,7 @@ const LinkPickerTitle = ({ title, type, description, onClear, onClick }) => (
 
 LinkPickerTitle.propTypes = {
   title: PropTypes.string.isRequired,
-  type: LinkType,
+  linkTypeTitle: PropTypes.string,
```
maxime-rainville: Doesn't match the `typeTitle` prop name above.

For context, we are definitely going to add link type icons down the track. And maybe other niceties like extended link type description. I think we should keep the Link Type as an object here.

client/src/entwine/JsonField.js
```diff
@@ -30,13 +30,12 @@ jQuery.entwine('ss', ($) => {
       const props = this.getProps();
       const ReactField = this.getComponent();
       const Root = this.getRoot();
-      Root.render(<ReactField {...props}  noHolder/>);
+      Root.render(<ReactField {...props} noHolder/>);
     },
 
-    handleChange(event, {id, value}) {
+    handleChange(linkID) {
```
maxime-rainville: That's not going to work in Redux form and elemental.

You need to retain this signature for your `onChange` handler.

client/src/entwine/JsonField.js
```diff
@@ -30,13 +30,12 @@ jQuery.entwine('ss', ($) => {
...
       const fieldID = $(this).data('field-id');
-      $('#' + fieldID).val(JSON.stringify(value)).trigger('change');
-      this.refresh();
+      $('#' + fieldID).val(JSON.stringify({ID: linkID}));
```
maxime-rainville: If we're only storing the link ID, why put it in the object.

client/src/entwine/JsonField.js
```diff
@@ -30,13 +30,12 @@ jQuery.entwine('ss', ($) => {
...
       const Root = this.getRoot();
-      Root.render(<ReactField {...props}  noHolder/>);
+      Root.render(<ReactField {...props} noHolder/>);
     },
 
-    handleChange(event, {id, value}) {
+    handleChange(linkID) {
       const fieldID = $(this).data('field-id');
-      $('#' + fieldID).val(JSON.stringify(value)).trigger('change');
-      this.refresh();
```
maxime-rainville, line 39: You need to keep calling this thing here to tell you component to update with its new props.

docs/en/linkable-migration.md
maxime-rainville, line 1: Why are we removing this? I'm pretty sure we will want to keep this migration path.

src/Controllers/LinkFieldController.php
```diff
@@ -0,0 +1,295 @@
...
+use SilverStripe\Security\SecurityToken;
+use SilverStripe\Forms\FieldList;
+use SilverStripe\Forms\FormAction;
+use SilverStripe\LinkField\Models\FileLink;
+use SilverStripe\ORM\ValidationException;
+use SilverStripe\ORM\ValidationResult;
+use SilverStripe\Control\Controller;
+use SilverStripe\Core\Injector\Injector;
+
+class LinkFieldController extends LeftAndMain
```
maxime-rainville, line 22: I'm OK with using this for now ... with the assumption that we will have a better way of creating API end points down the track.

src/Controllers/LinkFieldController.php
```diff
@@ -0,0 +1,295 @@
...
+
+    private static $allowed_actions = [
+        'linkForm',
+        'linkData',
+        'linkDelete',
+    ];
+
+    public function getClientConfig()
+    {
+        $clientConfig = parent::getClientConfig();
```
maxime-rainville, line 42: Presumably, we would only need these other client config if we were serving an actual UI through this controller?

src/Controllers/LinkFieldController.php
```diff
@@ -0,0 +1,295 @@
...
+            $typeKey = $this->typeKeyFromRequest();
+            $link = Registry::create()->byKey($typeKey);
+            if (!$link) {
+                $this->jsonError(400, 'Invalid typeKey');
+            }
+        }
+        if (!$link->canView()) {
+            $this->jsonError(403, 'Unauthorized');
+        }
+        return $this->createLinkForm($link);
```
maxime-rainville: Shouldn't we making the form ReadOnly the user doesn't have Write permission.

We should check CanCreate as well.

src/Controllers/LinkFieldController.php
maxime-rainville: I'm happy to split off the GraphQL parts in follow up cards so we can get the Link creation/update part over the line.

src/Controllers/LinkFieldController.php
maxime-rainville: I'm happy to split off the GraphQL parts in follow up cards so we can get the Link creation/update part over the line.

src/Controllers/LinkFieldController.php
```diff
@@ -0,0 +1,295 @@
...
+     * The 'save' method is called because it is the FormAction set on the Form
+     */
+    public function save(array $data, Form $form): HTTPResponse
+    {
+        if (empty($data)) {
+            $this->jsonError(400, 'Empty data');
+        }
+        // Ensure that ItemID url param is used as the source of truth for ID. ID shouldn't be in the
+        // POST request, though just remove it in case someone tries to add it in
+        if (isset($data['ID'])) {
```
maxime-rainville: Shouldn't we be throwing an exception if the ID from the form field doesn't match the ID in our URL parameter?

src/Controllers/LinkFieldController.php
```diff
@@ -0,0 +1,295 @@
...
+            unset($data['ID']);
+        }
+
+        /** @var Link $link */
+        $id = (int) $this->itemIDFromRequest();
+        if ($id) {
+            // Editing an existing Link
+            $link = Link::get()->byID($id);
+            if (!$link) {
+                $this->jsonErorr(400, 'Invalid ID');
```
maxime-rainville: I don't have strong feelings either way, but shouldn't this be a 404?

src/Controllers/LinkFieldController.php
```diff
@@ -0,0 +1,295 @@
...
+
+    /**
+     * Arrive here from FormRequestHandler::httpSubmission() during a POST request to
+     * /admin/linkfield/linkForm/<LinkID>
+     * The 'save' method is called because it is the FormAction set on the Form
+     */
+    public function save(array $data, Form $form): HTTPResponse
+    {
+        if (empty($data)) {
+            $this->jsonError(400, 'Empty data');
```
maxime-rainville, line 141: I'm assuming none of these error messages will be visible to the user, so they don't need to be translated?

src/Controllers/LinkFieldController.php
```diff
@@ -0,0 +1,295 @@
...
+            // $form->saveInto($link); doesn't seem to handle this
+            $link->FileID = $data['File']['Files'][0] ?? 0;
+        }
+
+        // DataObject validation
+        // thrown ValidationException will be caught in FormRequestHandler::httpSubmission()
+        // Note: Form (as opposed to DataObject) validate() is run in FormRequestHandler::httpSubmission()
+        $validationResult = $link->validate();
+        if (!$validationResult->isValid()) {
+            throw new ValidationException($validationResult);
```
maxime-rainville:
```suggestion
            throw ValidationException::create($validationResult);
```

src/Controllers/LinkFieldController.php
```diff
@@ -0,0 +1,295 @@
...
+        $id = $link->ID;
+
+        // Create the form
+        $formFactory = Injector::inst()->get(DefaultFormFactory::class);
+        $name = sprintf(self::FORM_NAME_TEMPLATE, $id);
+        /** @var Form $form */
+        $form = $formFactory->getForm($this, $name, ['Record' => $link]);
+
+        // Set where the form is submitted to
+        // Ideally would use getClientConfig() 'saveUrl' instaed, not sure why that doesn't work
```
maxime-rainville:
```suggestion
        // Ideally would use getClientConfig() 'saveUrl' instead, not sure why that doesn't work
```

src/Controllers/LinkFieldController.php
```diff
@@ -0,0 +1,295 @@
...
+        $urlSegment = $this->config()->get('url_segment');
+        $typeKey = Registry::create()->keyByClassName($link->ClassName);
+        $form->setFormAction("admin/$urlSegment/linkForm/$id?typeKey=$typeKey");
```
maxime-rainville: If you customise your admin area to use a URL other than `admin` that will break.

src/Controllers/LinkFieldController.php
```diff
@@ -0,0 +1,295 @@
...
+
+        // Add save action button
+        $title = $id ? 'Update link' : 'Create link'; // todo: _t()
```
maxime-rainville, line 229: Translate this.

src/Controllers/LinkFieldController.php
```diff
@@ -0,0 +1,295 @@
...
+     */
+    private function linkFromRequest(): Link
+    {
+        $itemID = (int) $this->itemIDFromRequest();
+        if (!$itemID) {
+            $this->jsonError(400, 'Invalid ID');
+        }
+        $link = Link::get()->byID($itemID);
+        if (!$link) {
+            $this->jsonError(400, 'Invalid ID');
```
maxime-rainville: If we can't find the link, shouldn't be a 404?

I'm not overly fuss about this.

src/Form/JsonField.php
```diff
@@ -40,37 +40,14 @@ public function saveInto(DataObjectInterface $record)
...
-            $record->{$fieldname} = $value;
-        }
+        // This code will all change in a later PR once we refactor out the JSON data format
+        // This method used to update the Link DataObject which was a relation
+        // Update the Pages has_one relationship ID using the existing JSON data format
+        $data = $this->parseString($this->dataValue());
+        // Not 'ID' key is perfectly valid e.g. removing an existing link
+        $linkID = $data['ID'] ?? 0;
+        $dbColumn = $fieldname . 'ID';
+        $record->$dbColumn = $linkID;
```
maxime-rainville, line 45: Isn't there a risk that we will create orphan links? Example: You created your link, but didn't save the page.

Happy to spin this off in a separate card.

src/GraphQL/LinkTypeResolver.php
```diff
@@ -17,10 +17,9 @@ public static function resolve($obj, $args = [], $context = [], ?ResolveInfo $in
         }
 
         $types = Registry::singleton()->list();
-        $flattenType = array_map(function (Type $type, string $key) {
+        $flattenType = array_map(function (Link $type, string $key) {
             return [
                 'key' => $key,
-                'handlerName' => $type->LinkTypeHandlerName(),
```
maxime-rainville, line 23: We're going to need to retain the handlerName to keep the capability to create custom link modals.

src/Models/FileLink.php
```diff
@@ -18,28 +12,14 @@ class FileLink extends Link
...
-        }
-
-        $file = File::get()->byID($fileId);
-
-        return $file?->getFilename() ?? '';
-    }
-
-    public function LinkTypeHandlerName(): string
-    {
-        return 'InsertMediaModal';
```
maxime-rainville, line 36: We need to keep that somehow.

src/Models/Link.php
```diff
@@ -192,12 +178,9 @@ public function jsonSerialize(): mixed
             return [];
         }
 
+        // TODO: this could lead to data disclosure - we should only return the fields that are actually needed
         $data = $this->toMap();
         $data['typeKey'] = $typeKey;
-        // Some of our models (SiteTreeLink in particular) have defined getTitle() methods. We *don't* want to override
-        // the 'Title' field (which represent the literal 'Title' Database field) - if we did that, then it would also
-        // apply this value into our Edit form. This addition is only use in the LinkField summary
-        $data['TitleRelField'] = $this->relField('Title');
```
maxime-rainville, line 200: We should have a follow up card so that the link title can be inferred from it's metadata later where appropriate. e.g.: Pointing a link to a Page probably means that the page title would be a suitable title.

src/Tasks/LinkableMigrationTask.php
maxime-rainville, line 1: I want to retain this for now.

src/Type/Registry.php
maxime-rainville, line 1: This class will be going away in a follow up card. But happy to keep it around for now.

tests/php/Models/LinkTest.php
maxime-rainville: I'm surprised that we even need a physical file for this. I would have thought that a File entry in the DB would have been sufficient.

maxime-rainville (changes requested):
This is just an initial review. I'm planning on having a second pass.

- The re-factored link field doesn't work at all in Elemental.
- This is probably because the Root state management has some problems.
- We've lost the FileLinkModal. I'm fine with spinning this off into it's own card because it is a bit of a headache to get working correctly, but I would like us to at least be confident the new approach can support it.

_config/config.yml
maxime-rainville: They do.

client/src/components/LinkPicker/LinkPicker.js
ChloeSartorelli: `!!` is a code smell imo. It's fine if it ends up that way in the dist files but we should aim for no ambiguity in source code.

src/Models/FileLink.php
emteknetnz: We're using FormSchema/FormBuilder for everything now, so the FormBuilderModal is the only viable option. How come we need InsertMediaModal? FileLink works correctly on my local.

src/Controllers/LinkFieldController.php
emteknetnz: Correct, you only see them in browser developer tools, not in a toast

src/Form/JsonField.php
emteknetnz: Yes that's a possibility, it's just the nature of saving child DataObjects independently of the page save rather than all at once on page save.

The user will at least get a "there are unsaved changes on this page" if you try to navigate away from the page without saving.

src/GraphQL/LinkTypeResolver.php
emteknetnz: I'm not sure if that is viable anymore since we have a system that relies on FormSchema which FormBuilderModal seems like the only option

Even if it were easily technically doable, I'm not sure that it's a problem that even needs to be solved? What's an actual real world use case where you need a different modal to manage links that FormBuilderModal cannot handle?

client/src/components/LinkPicker/LinkPicker.js
maxime-rainville: You can use `Boolean(typeTitle)` as well.

It's a very minor thing.

src/Form/JsonField.php
maxime-rainville: I think we should try to assign the owner at the same time we do the initial save, we can look at this once we pick up reverse ownership again.

src/GraphQL/LinkTypeResolver.php
maxime-rainville: We have an immediate use case for creating link for File. The FileInsertModal is design so you can use different form schema inside of it:
- Insert a link to file in the WYSIWYG,
- Insert an image in the in the WYSIWYG,
- Edit file details,
- Edit folder details.


I could imagine complex scenarios where you might want to display a complex form for creating links:
- I want to create a link to GitHub issues and I need to build a search form in my modal so I can get there.

client/src/components/LinkField/LinkField.js
```diff
@@ -1,90 +1,141 @@
...
+ * actions - object of redux actions
+ */
+const LinkField = ({ value, onChange, types, actions }) => {
+  const [typeKey, setTypeKey] = useState(value.typeKey || '');
+  const [linkID, setLinkID] = useState(value.ID || 0);
+  const [data, setData] = useState(value);
+  const [editing, setEditing] = useState(false);
 
-const LinkField = ({ id, loading, Loading, data, LinkPicker, onChange, types, linkDescription, ...props }) => {
-  if (loading) {
```
maxime-rainville, line 7: We should retain the loading logic. There will be time where we are waiting for information to came back from the server.

client/src/components/LinkField/LinkField.js
```diff
@@ -1,90 +1,141 @@
...
 
-  const onClear = (event) => {
-    if (typeof onChange !== 'function') {
-      return;
+      // success toast
+      actions.toasts.success('Saved link');
     }
 
-    onChange(event, { id, value: {} });
+    return Promise.resolve();
```
maxime-rainville, line 55: We're mixing promises and await/async. Do we want to normalise everything to use the same syntax?

client/src/components/LinkField/LinkField.js
maxime-rainville: Can we look at moving all the methods that do HTTP request to their own file?

It's somewhat manageable now, but once we throw in Link lists in there I think it's going to became a bit more difficult.

Having them as individual method will also make it easier to write JEST test for them.

maxime-rainville (changes requested):
I had second pass. I think the approach looks broadly solid.

I'll create some follow up cards for problems that I don't think need to be addressed in this card.

Some key things I think need to be addressed now:
- The link field needs to work in Elemental
- Root state management
- Retain the ability to use different Link modal (I'm fine if the file link modal is broken for now)
- Do we want to store just the ID in the parent input?
- Making sure LinkField still work if you customise your `admin` area url.

Things that need to be addressed, but can be spun off as related card
- [Handling all possible errors form AJAX requests](https://github.com/silverstripe/silverstripe-linkfield/issues/110)
- [Loading UI](https://github.com/silverstripe/silverstripe-linkfield/issues/111)
- [File link modal and alternative link modal](https://github.com/silverstripe/silverstripe-linkfield/issues/112)
- Using GraphQL to for initial link description and clearing link
- [Localisation](https://github.com/silverstripe/silverstripe-linkfield/issues/113)
- Controller for parking API end points and form schema
- Readonly state of form schema

Other things that we can decide later
- Migration from other modules ... just retain the existing one for now.
- Orphan links.
- Inferring Link title 


FYI I didn't look at the fine print of all the test coverage for now, since I presume there's a substantial bit that still needs to be updated.

src/Form/JsonField.php
emteknetnz: I don't think this is possible if as we're using has_one relations e.g. MyPage.has_one = [MyLink => Link::class].  Reason being that we'd need to call $page->write() in order to save the relation since the foriegn key is on the Page, not the Link.  It's simply wrong to do this in an XHR request to save a Link, the correct time is when you click the save button on the Page edit form.

src/GraphQL/LinkTypeResolver.php
emteknetnz: > We have an immediate use case for creating link for File

... FileLink works in this PR using the FormBuilderModal, is it not work on your local?

Using the FormBuilderModal it renders the File relation as a regular UploadField (consistent with Page edit form). It's possibly better than before previously the InsertMediaModal the Link.Title field wasn't being shown on the Link edit form (because there wasn't a form), now it is.

client/src/components/LinkField/LinkField.js
emteknetnz: Not sure what quite sure why we'd move HTTP request stuff to a different file. HTTP request are:
- GraphQL query to get link types
- fetch() (lib/Backend) query to fetch data
- fetch() (lib/Backend) query to delete link

The queries are simple and don't take up much in the file. Splitting things across files seems like it's needlessly increasing complexity.

For the 2x fetch() requests we can mock in jest them pretty easily - e.g. https://github.com/silverstripe/silverstripe-session-manager/blob/2/client/src/components/LoginSession/tests/LoginSessionContainer-test.js#L10

For GraphQL, I don't know how to mock the responses in jest

client/src/components/LinkField/LinkField.js
emteknetnz: There's a convoluted stack of callback handler being passed around LinksField.js, [FormBuilderLoader.js](https://github.com/silverstripe/silverstripe-admin/blob/2/client/src/containers/FormBuilderLoader/FormBuilderLoader.js#L97) and [FormBuilder.js](https://github.com/silverstripe/silverstripe-admin/blob/2/client/src/components/FormBuilder/FormBuilder.js#L237) (maybe others?) What we have now works and I don't want change it

src/Tasks/LinkableMigrationTask.php
emteknetnz: Reverted

client/src/components/LinkField/LinkField.js
emteknetnz: Updated

client/src/components/LinkField/LinkField.js
emteknetnz: Keeping the setData({}) as data is managed on the LinkField. Have removed the setLinkID(0), as managing the LinkID on the parent JsonField.

client/src/components/LinkPicker/LinkPicker.js
emteknetnz: Will just leave as is

client/src/components/LinkPicker/LinkPicker.js
emteknetnz: Updated

client/src/components/LinkPicker/LinkPickerTitle.js
emteknetnz: Updated to linkType.

We can update the prop / propType in a later PR when we add icons, shouldn't do it now because they're not supported

client/src/components/LinkPicker/LinkPickerTitle.js

client/src/entwine/JsonField.js
emteknetnz: Updated

client/src/entwine/JsonField.js
emteknetnz: Updated

src/Controllers/LinkFieldController.php
emteknetnz: It's standard to merge the config with parent::getClientConfig() in other places in the codebase where we define getClientConfig();

src/Controllers/LinkFieldController.php
emteknetnz: Updated to check for these and set to readonly in createLinkForm()

src/Controllers/LinkFieldController.php

src/Controllers/LinkFieldController.php
emteknetnz: Updated all of the 'Invalid ID' / 'Invalid typeKey' errors to 404

src/Controllers/LinkFieldController.php

src/Controllers/LinkFieldController.php

emteknetnz:
@maxime-rainville I've split off a separate card for elemental support, looks like it's non-trivial - https://github.com/silverstripe/silverstripe-linkfield/issues/116

src/Controllers/LinkFieldController.php

client/src/entwine/JsonField.js
emteknetnz: I don't understand why this is relevant for redux-form or elemental. The handleChange/onChange callback is only used to communicate between JsonField and LinkField, it's not passed down to any child components

client/src/components/LinkField/LinkField.js
emteknetnz: Will do in https://github.com/silverstripe/silverstripe-linkfield/issues/113

client/src/components/LinkField/LinkField.js
emteknetnz: Will do in https://github.com/silverstripe/silverstripe-linkfield/issues/113

client/src/components/LinkField/LinkField.js
emteknetnz: Will do in https://github.com/silverstripe/silverstripe-linkfield/issues/113

client/src/components/LinkField/LinkField.js
emteknetnz: It's currently a one liner, I don't think it needs to be optimised any further

src/Controllers/LinkFieldController.php
emteknetnz: Will do in https://github.com/silverstripe/silverstripe-linkfield/issues/113

src/GraphQL/LinkTypeResolver.php
emteknetnz: Have reverted

src/Models/FileLink.php
emteknetnz: Can look at reimplementing in https://github.com/silverstripe/silverstripe-linkfield/issues/112

client/src/components/LinkField/LinkField.js
emteknetnz: Can reimplement in https://github.com/silverstripe/silverstripe-linkfield/issues/111

client/src/components/LinkField/LinkField.js
emteknetnz: Can implement in https://github.com/silverstripe/silverstripe-linkfield/issues/110

client/src/entwine/JsonField.js
maxime-rainville: The redux onChange handler expect a different method signature.

If we have one method signature for entwine and one for redux form, the field will have to be aware of how its being rendered and uddate the return value accordingly.

Much easier to just have the entwine unchange handler expect the same thing as redux form.

client/src/components/LinkField/LinkField.js
maxime-rainville: Having played with the thing a bit more, the `const match = formSchema.id.match(/\/linkForm\/([0-9]+)/)` logic doesn't seem robust.

I'm thinking this could be substantially better.

- If you're form validation that should return a 400 along with a form schema.
- If you're form submission is successful, then it shouldn't return a form schema at all. It should return the ID of the successfully created record. That would mimic tho behaviour of a regular Form object where we return a view with a success message.

client/src/components/LinkField/LinkField.js
```diff
@@ -1,90 +1,148 @@
...
+    <LinkModal {...modalProps} />
+  </>;
+};
 
-  return <Fragment>
-      <LinkPicker {...linkProps} />
-      <LinkModal {...modalProps} />
-    </Fragment>;
+LinkField.propTypes = {
+  linkID: PropTypes.number.isRequired,
```
maxime-rainville: I think we need to agree with some convention up front.

`linkID` should be called `value`. This is not just an academic argument. Redux-form will call it `value` when it passes it in the form schema.

Equally, `id` should be reserved for the id of the form field.

client/src/components/LinkField/LinkField.js
maxime-rainville: Looking at the fine print of this, when we're in the process of creating a new entry, it uses an id of `0` for the URL. `/admin/linkfield/schema/linkForm/0?typeKey=phone`

I'm wondering if we should use something like `new` or maybe nothing at all?

If we want to be REST purists, we could say that the endpoint for updating an existing link should be a `PUT`. Creation would occur on `POST`. 

Probably doesn't matter on the grander scheme of things. But if we want this to be the basis for a more generic approach, it's probably worth sorting it now.

client/src/components/LinkPicker/LinkPickerTitle.js
maxime-rainville: If it was already like that I would be fine with it. But we are literally removing something we are planning to bring back in a not too distant PR.

client/src/entwine/JsonField.js
emteknetnz: I don't see how this is easier? I'm really confused as to how redux-form is relevant here. This handler is only used between JsonField and LinkField

> the field will have to be aware of how its being rendered

By that do you mean the parent component on LinkField is, i.e it could be something that is NOT JsonField?

I can kind of see how passing `event` as the first param makes sense, since that's fairly standard for event handlers, however in the context on onModalSubmit() we don't actually have an event passed down as an argument.  I think the code that consumes onModalSubmit() is https://github.com/silverstripe/silverstripe-admin/blob/2/client/src/components/FormBuilder/FormBuilder.js#L249

src/Controllers/LinkFieldController.php
```diff
@@ -0,0 +1,302 @@
...
+        $formFactory = Injector::inst()->get(DefaultFormFactory::class);
+        $name = sprintf(self::FORM_NAME_TEMPLATE, $id);
+        /** @var Form $form */
+        $form = $formFactory->getForm($this, $name, ['Record' => $link]);
+
+        // Set where the form is submitted to
+        $adminRoot = AdminRootController::get_admin_route();
+        $urlSegment = $this->config()->get('url_segment');
+        $typeKey = Registry::create()->keyByClassName($link->ClassName);
+        $form->setFormAction("$adminRoot/$urlSegment/linkForm/$id?typeKey=$typeKey");
```
maxime-rainville:
```suggestion
        $typeKey = Registry::create()->keyByClassName($link->ClassName);
        $form->setFormAction($this->Link("linkForm/$id?typeKey=$typeKey"));
```

maxime-rainville (changes requested):
>  I've split off a separate card for elemental support, looks like it's non-trivial

Making sure this new approach works with Elemental is pretty fundamental. I think getting it working in Elemental is mostly a question of stealing some bits from my `silverstripe-react` module.

I had a go to try to get it working in this PR and mostly got there. I'm confident the new approach can be made to work with the new form schema approach.

It took me a couple days to wrap my head around getting a form field working in React and Entwine. Probably best if I pick the elemental follow up card.

client/src/components/LinkField/LinkField.js
emteknetnz: > If we want to be REST purists, we could say that the endpoint for updating an existing link should be a PUT. Creation would occur on POST.

FormSchema is not REST :-) e.g. we POST x-form-urlencoded data to the endpoint, not application/json.  FormSchema is its own weird little standard.  If we try to make this more RESTful then we'll just end up doing a bunch of stuff to get it more compliant, little of which has any real world benefit IMO. Let's not even bother to try to get this more RESTful.

>> I'm wondering if we should use something like new or maybe nothing at all?

`/admin/linkfield/schema/linkForm?typeKey=phone` without the zero for new records would make more sense.  It'll work the same as zero as we just test for a falsey $ItemID request param which should be null instead of 0

client/src/components/LinkField/LinkField.js
emteknetnz: >> If you're form validation [failed] that should return a 400 along with a form schema.

I can't find the exact commit (it's been forced pushed over so many times), though I do remember there was a technical reason why I chose to not return a 400 on validation failure, which I'd definitely prefer to do.

From memory it's because in lib/Backend (Silverstripe wrapper for fetch()) we have some custom code to throw Errors on 400+ status codes (i.e. !response.ok) so that we can then use fetch's  `.catch()` to handle these.  However there's [this code in FormBuilder](https://github.com/silverstripe/silverstripe-admin/blob/2/client/src/components/FormBuilder/FormBuilder.js#L245) that will then just throw these Errors again and won't get the FormSchema returned, meaning that now validation messages don't show in the FormBuilder form

You can test this by updating the last bit of `$form->setValidationResponseCallback` in `LinkFieldController::createLinkForm()`

```php
            $response = $this->getSchemaResponse($schemaId, $form, $errors);
            $response->setStatusCode(400);
            return $response;
```

You'll get a 400 returned on validation failure, however you'll also get javascript errors in FormBuilderModal.js and Backend.js in browser developer tools

emteknetnz: > If you're form submission is successful, then it shouldn't return a form schema at all. It should return the ID of the successfully created record.

I don't agree.  To the best of my knowledge the existing FormBuilder/FormSchema "standard" is designed to expect a FormSchema with no errors returned on success. Presumably under the hood it just re-renders the form at that point using the returned FormSchema? Having a quick look at campaign admin it returns a FormSchema on success.

We absolutely should NOT be tinkering with FormBuilder/FormSchema at this stage IMO. While it's a kind of weird standard it does work pretty great and we should just implement it as is

client/src/components/LinkField/LinkField.js
emteknetnz: As mentioned in standup, will leave as is for now, we switch this out to something different later on before tagging stable if we want

client/src/components/LinkField/LinkField.js
emteknetnz: I've changed the prop to value which does make sense as it is the value of the FormField, though I've also kept a local variable `const linkID = value`; - it's simply a lot easier to follow the code as linkID is a more descriptive name

src/Controllers/LinkFieldController.php

client/src/components/LinkField/LinkField.js
maxime-rainville: Looking at Campaign Admin, on success it it's stick a `message` attribute on success that then gets displayed on the front end.

I don't like that behaviour. It should be displaying a toast instead.

I've added a few ACs to https://github.com/silverstripe/silverstripe-admin/issues/1589:
> - There's a standardised and documented way of creating new FormSchema endpoints in the CMS.
>   - This accounts for form validation
>   - *This accounts for communicating success to the front end (e.g. Displaying success message)*
>   - *Extra context about the success of the form submission can be communicated to the parent JS process*

Created a follow up card.

image
image

_graphql/queries.yml
maxime-rainville: https://github.com/silverstripe/silverstripe-linkfield/issues/15

client/src/components/LinkField/LinkField.js
maxime-rainville: Promises and await/async are the same thing under the hood. 

More senior JS dev will understand this, but junior dev might get confused.

That might be a wider discussion about our JS coding standards.

maxime-rainville (approved):
Not quite happy with everything. But happy enough to merge in the interest of keeping things moving and unblocking follow up cards.

PR #109, opened 2023-10-26
make translatable & add german translations

bumbus:


src/Models/EmailLink.php
```diff
@@ -36,4 +42,11 @@ public function getURL(): string
     {
         return $this->Email ? sprintf('mailto:%s', $this->Email) : '';
     }
+
+    public function fieldLabels($includerelations = true)
```
ChloeSartorelli, line 46: This isn't necessary - there's a fallback to a dB namespaced translatable key which you'll surface in the yml file if you run the i18n text collector.

Same with all other field labels you're adding

client/lang/de.js
```diff
@@ -0,0 +1,13 @@
+// This file was generated by silverstripe/cow from client/lang/src/en.json.
+// See https://github.com/tractorcow/cow for details
```
ChloeSartorelli: Please remove this false comment

GuySartorelli:
Thanks for your interest in this module and for the contribution.
Just an FYI, this module is still undergoing active development and we'll likely make this fully translatable as part of that development. As a result it's likely that we'll do that work separately and that work will supercede this pull request.

src/Models/EmailLink.php
bumbus: Hmm am I missing something, if I comment out `fieldLabels` function the translations are not applied.

Bildschirmfoto 2023-10-30 um 16 28 16

ChloeSartorelli: Did you run the i18n text collector over this module, and use the new translation keys it generates for your german translations?

GuySartorelli:
Closing in favour of https://github.com/silverstripe/silverstripe-linkfield/pull/119

Once v4 is closer to release we'll push the strings to transifex and you can add your German translations there.

Issue #113, opened 2023-11-02
LinkField can be translated


maxime-rainville:
## Story

As a content author, I want the ability to use LinkField in my native language.

## Acceptance criteria
- [x] Module is added to Transifex.
- [x] All content string in JS and PHP are localisable.
- [x] All field labels for all Link models are translatable.

## Exclude 
- Initial string to translate are uploaded to Transifex. We want to wait until the keys are final.

## PR
- https://github.com/silverstripe/silverstripe-linkfield/pull/119

emteknetnz:
Linked PR has been merged

PR #119, opened 2023-11-13
ENH Add localisation support

sabina-talipova:
## Description
Added localisation support. All available messages, field names, etc. are updated and have the value set in the localisation file

## Parent issue
- https://github.com/silverstripe/silverstripe-linkfield/issues/113


lang/en.yml
```diff
@@ -0,0 +1,34 @@
+en:
+  LinkField:
+    UPDATELINK: 'Update link'
```
emteknetnz: There isn't a clear standard around translation keys. While uppercase without underscores is the most common, is also hard to read

Could you please put underscores in the keys e.g. `UPDATE_LINK`

src/Models/ExternalLink.php
```diff
@@ -15,6 +17,15 @@ class ExternalLink extends Link
         'ExternalUrl' => 'Varchar',
     ];
 
+    public function getCMSFields(): FieldList
+    {
+        $this->beforeUpdateCMSFields(static function (FieldList $fields) {
```
emteknetnz: Could you please change `static function` to just `function` for all of the anonymous methods in the getCMSFields for all of the subclasses on Link

emteknetnz (changes requested):

lang/en.yml
sabina-talipova: Updated

src/Models/ExternalLink.php
sabina-talipova: Updated

emteknetnz (approved):

Issue #107, opened 2023-10-24
Multiple link support


emteknetnz:
## Story

As a CMS User I want to be able to add multiple links to a link field so that I am not limited to adding only a single link

## Acceptance criteria
- Ability to add multiple links to a link field
- Ability to delete links from a link field
- Multi link field works in both Elemental inline form and regular pages forms.
- ~~Design has been validated with UX team~~
- Documentation is updated to mention how to manage links of list

## Notes
- There is an implementation in an old PR that may be useful - https://github.com/silverstripe/silverstripe-linkfield/pull/45
- We never validated how the LinkField should look when rendering many links
- The POC splits the Single Link and Many Link fields into two different form fields. We're not tied to that approach, although it probably makes sense.
- We have some pre-existing problems with fields that need to read their relation on their parent data object. This has cause us problems with the Upload field in Elemental in the past.

## Exclude 
- Ability to sort multiple links within a link field (will split as a separate card)

## Rough design - doesn't have to be exact
![image](https://github.com/silverstripe/silverstripe-linkfield/assets/1168676/ff0aecee-e2e0-47b4-9633-c319b55569c3)

## PRs
- https://github.com/silverstripe/silverstripe-linkfield/pull/120

emteknetnz:
Linked PR has been merged

PR #120, opened 2023-11-28
NEW Add MultiLinkField

GuySartorelli:
## Notes
- I've refactored some of the original components to avoid code duplication - most of this can be seen in the first commit.
- Looking at the existing unit test coverage, there's nothing particular about this PR that needs to have new unit tests. I suspect a majority of it will be covered by behat tests later on instead, which is a separate card.

## Issue
- https://github.com/silverstripe/silverstripe-linkfield/issues/107


client/src/entwine/LinkField.js
```diff
@@ -45,20 +45,29 @@ jQuery.entwine('ss', ($) => {
...
+
     /**
      * Remove the component when unmatching
      */
     onunmatch() {
       const Root = this.getRoot();
-      Root.unmount();
+      if (Root) {
+        Root.unmount();
+      }
```
ChloeSartorelli, line 71: I was sometimes getting console errors about unmount not being a function on undefined.
See https://docs.silverstripe.org/en/5/changelogs/5.0.0/#reactdom-render-replaced-with-reactdom-createroot-render where this pattern of checking for the root before unmounting is documented as best practice anyway.

babel.config.json
ChloeSartorelli, line 1: These presets are already defined in the global webpack config.
See https://github.com/silverstripe/webpack-config/blob/5832637ef465be20afd50407cdae0185880166b5/js/modules.js#L15-L25

.eslintrc.js
```diff
@@ -1 +1,7 @@
-module.exports = require('@silverstripe/eslint-config/.eslintrc');
+module.exports = {
+  extends: '@silverstripe/eslint-config',
+  // Allows null coalescing and optional chaining operators.
+  parserOptions: {
+    ecmaVersion: 2020
+  },
```
ChloeSartorelli, line 6: Null coallescing (`x ?? y`) and optional chaining (`x?.y`) are well supported now - the build would have failed if they weren't supported by the browsers we target, due to the browser config in package.json

client/src/components/LinkField/LinkField.js
ChloeSartorelli, line 1: This component now handles single and multi link fields.
I originally had them as two separate components, but the differences were so minimal it just made more sense to have a `isMulti` boolean prop and let the one component handle both scenarios.

Note that I've moved the `LinkPickerTitle` out of the picker, so that it can be used to render the links when linkfield is multi.

The picker is now responsible for knowing what type was picked, and passing it through to the modal. The linkfield shouldn't care about that, it only cares about the link once the link has been created.

I've also moved the responsibility of rendering the modal. If it's a new link, the modal is rendered by the picker (indirectly, through the new `LinkModalContainer`). The linkfield itself only renders modals for links that are being edited.

Finally, the logic for figuring out _which_ link modal component to render was moved into `LinkModalContainer` so it didn't get duplicated between the linkfield and the picker components.

client/src/components/LinkModal/LinkModal.js
```diff
@@ -17,13 +17,37 @@ const buildSchemaUrl = (typeKey, linkID) => {
...
-const LinkModal = ({ typeTitle, typeKey, linkID, editing, onSubmit, onClosed}) => {
+const LinkModal = ({ typeTitle, typeKey, linkID = 0, isOpen, onSuccess, onClosed}) => {
   if (!typeKey) {
     return false;
   }
+
+  /**
+   * Call back used by LinkModal after the form has been submitted and the response has been received
+   */
+  const onSubmit = async (modalData, action, submitFn) => {
```
ChloeSartorelli, line 28: Moved a bunch of this logic from linkfield into here - linkfield doesn't care that the modal was submitted, it only cares about success. Similarly, the modal should be responsible for knowing about validation errors - linkfield itself doesn't care about them.

client/src/components/LinkPicker/LinkPickerTitle.js
ChloeSartorelli, line 1: The only change to `LinkPickerTitle` was to give it the classnames necessary for it to look correct when rendered outside of the picker component.

package.json
```diff
@@ -49,8 +49,8 @@
   },
   "devDependencies": {
     "@babel/runtime": "^7.20.0",
-    "@silverstripe/eslint-config": "^1.0.0-alpha6",
-    "@silverstripe/webpack-config": "^2.0.0-alpha9",
+    "@silverstripe/eslint-config": "^1.0.0",
+    "@silverstripe/webpack-config": "^2.0.0",
```
ChloeSartorelli, line 53: Shouldn't be using alpha anymore.

client/src/components/LinkField/LinkField.js
```diff
@@ -87,69 +114,65 @@ const LinkField = ({ value, onChange, types, actions }) => {
...
+        onClosed={onModalClosed}
+        linkID={editingID}
+      />
+    }
   </>;
 };
 
 LinkField.propTypes = {
-  value: PropTypes.number.isRequired,
+  value: PropTypes.oneOfType([PropTypes.arrayOf(PropTypes.number), PropTypes.number]),
```
ChloeSartorelli, line 171: Can't be required, because the value is `undefined` when no links are in the field and the field is rendered in an elemental block.
The component handles `undefined` perfectly so that's not a problem.

client/src/components/LinkPicker/LinkPicker.scss
ChloeSartorelli, line 1: I've reduced all of the `&__` down to their full `.link-picker__` forms so that we can easily do a text search for any given CSS class and find the style here.

client/src/components/LinkPicker/LinkPickerMenu.js
```diff
@@ -15,7 +15,7 @@ const LinkPickerMenu = ({ types, onSelect }) => {
       toggle={toggle}
       className="link-picker__menu"
     >
-      <DropdownToggle className="link-picker__menu-toggle font-icon-link" caret>{i18n._t('LinkField.ADD_LINK', 'Add Link')}</DropdownToggle>
+      <DropdownToggle className="link-picker__menu-toggle font-icon-plus-1" caret>{i18n._t('LinkField.ADD_LINK', 'Add Link')}</DropdownToggle>
```
ChloeSartorelli, line 18: Now that we are rendering links below the picker, it makes way more sense to have a `+` symbol for the picker, and a link symbol for the links. It's a super quick visual distinction.

README.md
```diff
@@ -19,22 +19,10 @@ This module provides a Link model and CMS interface for managing different types
...
 
-### GraphQL v4 - Silverstripe 4
-
-`composer require silverstripe/linkfield:^2`
-
-### GraphQL v3 - Silverstripe 4
-
-```sh
-composer require silverstripe/linkfield:^1
-```
```
ChloeSartorelli, line 36: This branch is only valid for CMS 5

README.md
```diff
@@ -67,19 +63,11 @@ class Page extends SiteTree
 }
 ```
 
-## Migrating from Version `1.0.0` or `dev-master`
-
-Please be aware that in early versions of this module (and in untagged `dev-master`) there were no table names defined
-for our `Link` classes. These have now all been defined, which may mean that you need to rename your old tables, or
-migrate the data across.
-
-EG: `SilverStripe_LinkField_Models_Link` needs to be migrated to `LinkField_Link`.
```
ChloeSartorelli, line 76: These are old and don't apply to the jump from 2.x/3.x to 4.x

README.md
```diff
@@ -43,22 +31,30 @@ use SilverStripe\CMS\Model\SiteTree;
...
 
 class Page extends SiteTree
 {
     private static array $has_one = [
         'HasOneLink' => Link::class,
     ];
 
+    private static $many_many = [
+        'Links' => Link::class
+    ];
```
emteknetnz:
```suggestion
    private static $has_many = [
        'HasManyLinks' => Link::class
    ];
```
Should be has many since that's the normal use case (and I'm not sure we're even attempting to support many_many given ownership model). Doesn't matter that it won't work just yet.  Name field consistently with 'HasOneLink'

README.md
```diff
@@ -43,22 +31,30 @@ use SilverStripe\CMS\Model\SiteTree;
...
 
+        // Don't forget to remove the auto-scaffolded fields!
+        $fields->removeByName(['HasOneLinkID', 'Links']);
+
         $fields->addFieldsToTab(
             'Root.Main',
             [
                 LinkField::create('HasOneLink'),
-                LinkField::create('DbLink'),
+                MultiLinkField::create('Links'),
```
emteknetnz:
```suggestion
                MultiLinkField::create('HasManyLinks'),
```

client/src/components/LinkField/LinkField.js
```diff
@@ -19,36 +22,63 @@ const section = 'SilverStripe\\LinkField\\Controllers\\LinkFieldController';
...
+  // Ensure we have a valid array
+  let linkIDs = value;
+  if (!Array.isArray(linkIDs)) {
+    if (!linkIDs) {
+      linkIDs = [];
+    }
+    if (typeof linkIDs === 'number') {
+      linkIDs = [linkIDs];
+    }
+  }
```
emteknetnz:
```suggestion
  let linkIDs = value;
  if (!Array.isArray(linkIDs)) {
    linkIDs = [];
    if (typeof value === 'number') {
      linkIDs = [value];
    }
  }
```

client/src/components/LinkField/LinkField.js
```diff
@@ -19,36 +22,63 @@ const section = 'SilverStripe\\LinkField\\Controllers\\LinkFieldController';
...
+
+  // Read data from endpoint and update component state
+  // This happens any time a link is added or removed and triggers a re-render
+  useEffect(() => {
+    if (!editingID && linkIDs.length > 0) {
+      const query = [];
+      for (const id of linkIDs) {
+        query.push(`itemID[]=${id}`);
```
emteknetnz: This results in really strange looking requests
/admin/linkfield/data/?itemID[]=2&itemID[]=4 

Should be
/admin/linkfield/data/2,4

client/src/components/LinkField/LinkField.js
```diff
@@ -19,36 +22,63 @@ const section = 'SilverStripe\\LinkField\\Controllers\\LinkFieldController';
...
+      }
+      const endpoint = `${Config.getSection(section).form.linkForm.dataUrl}/?${query.join('&')}`;
+      backend.get(endpoint)
+        .then(response => response.json())
+        .then(responseJson => {
+          setData(responseJson);
+        });
+    }
+  }, [editingID, value && value.length]);
```
emteknetnz, line 57:
```suggestion
  }, [editingID]);
```
There's already a conditional to check linkIDs.length within the useEffect block

client/src/components/LinkField/LinkField.js
```diff
@@ -19,36 +22,63 @@ const section = 'SilverStripe\\LinkField\\Controllers\\LinkFieldController';
...
+   */
+  const onModalSuccess = (value) => {
       // update component state
-      setEditing(false);
+      setEditingID(null);
 
-      // update parent JsonField data id - this is required to update the underlying <input> form field
-      // so that the Page (or other parent DataObject) gets the Link relation ID set
-      onChange(valueFromSchemaResponse);
+      let ids = [...linkIDs];
```
emteknetnz:
```suggestion
      const ids = [...linkIDs];
```

client/src/components/LinkField/LinkField.js
```diff
@@ -57,16 +87,13 @@ const LinkField = ({ value, onChange, types, actions }) => {
...
+  }
 
   /**
-   * Call back used by LinkPicker when the 'Clear' button is clicked
+   * Update the component when the 'Clear' button in the LinkPicker is clicked
    */
-  const onClear = () => {
-    const endpoint = `${Config.getSection(section).form.linkForm.deleteUrl}/${linkID}`;
+  const onClear = (id) => {
+    const endpoint = `${Config.getSection(section).form.linkForm.deleteUrl}/${id}`;
```
emteknetnz:
```suggestion
  const onClear = (linkID) => {
    const endpoint = `${Config.getSection(section).form.linkForm.deleteUrl}/${linkID}`;
```

client/src/components/LinkPicker/LinkPicker.js
```diff
@@ -1,31 +1,63 @@
...
+   * When a link type is selected, set the type key so we can open the modal.
+   */
+  const doSelect = (key) => {
+    setTypeKey(key);
+  }
+
+  /**
+   * Callback for when the modal is closed by the user
+   */
+  const onClosed = () => {
```
emteknetnz:
```suggestion
  const handleClosed = () => {
```
React convention

client/src/components/LinkPicker/LinkPicker.js
```diff
@@ -1,31 +1,63 @@
...
+    if (typeof onModalClosed === 'function') {
+      onModalClosed();
+    }
+    setTypeKey('');
+  }
+
+  /**
+   * Callback for when the modal successfully saves a link
+   */
+  const onSuccess = (value) => {
```
emteknetnz:
```suggestion
  const handleSuccess = (value) => {
```

client/src/components/LinkPicker/LinkPicker.js
```diff
@@ -1,31 +1,63 @@
...
+
+  return (
+    <div className={classnames('link-picker', 'form-control')}>
+      <LinkPickerMenu types={Object.values(types)} onSelect={doSelect} />
+      { shouldOpenModal && <LinkModalContainer
+          types={types}
+          typeKey={typeKey}
+          isOpen={shouldOpenModal}
+          onSuccess={onSuccess}
+          onClosed={onClosed}
```
emteknetnz:
```suggestion
          onSuccess={handleSuccess}
          onClosed={handleClosed}
```

client/src/components/LinkPicker/LinkPicker.js
```diff
@@ -1,31 +1,63 @@
...
+/**
+ * Component which allows users to choose a type of link to create, and opens a modal form for it.
+ */
+const LinkPicker = ({ types, onModalSuccess, onModalClosed }) => {
+  const [typeKey, setTypeKey] = useState('');
+
+  /**
+   * When a link type is selected, set the type key so we can open the modal.
+   */
+  const doSelect = (key) => {
```
emteknetnz:
```suggestion
  const handleSelect = (key) => {
```

client/src/components/LinkPicker/LinkPicker.js
```diff
@@ -1,31 +1,63 @@
...
+
+  /**
+   * Callback for when the modal successfully saves a link
+   */
+  const onSuccess = (value) => {
+    setTypeKey('');
+    onModalSuccess(value);
+  }
+
+  const shouldOpenModal = Boolean(typeKey);
```
emteknetnz:
```suggestion
  const shouldOpenModal = typeKey !== '';
```
Save having to go up and check cases when this would be true/false

client/src/components/LinkPicker/LinkPicker.js
```diff
@@ -1,31 +1,63 @@
...
+
+  return (
+    <div className={classnames('link-picker', 'form-control')}>
+      <LinkPickerMenu types={Object.values(types)} onSelect={doSelect} />
```
emteknetnz:
```suggestion
const types = Object.values(types);
const className = classnames('link-picker', 'form-control');

  return (
    <div className={className}>
      <LinkPickerMenu types={types} onSelect={handleSelect}/>
```

A bit easier to read

client/src/components/LinkPicker/LinkPicker.scss
```diff
@@ -1,79 +1,94 @@
-.link-picker {
+%link-row {
```
emteknetnz, line 1: What is the % sign? Should be dot instead?

client/src/containers/LinkModalContainer.js
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Contains the LinkModal and determines which modal component to render based on the link type.
+ */
+const LinkModalContainer = ({ types, typeKey, linkID = 0, isOpen, onSuccess, onClosed}) => {
+  if (!typeKey) {
+    return false;
+  }
+
+  const type = types.hasOwnProperty(typeKey) ? types[typeKey] : {};
```
emteknetnz, line 15: How do things work if typeKey is not in types?  Should we just return false if that happens?

src/Controllers/LinkFieldController.php
```diff
@@ -274,17 +291,56 @@ private function linkFromRequest(): Link
...
+        return (int) $itemID;
+    }
+
+    /**
+     * Get the value of the itemID request query string argument
+     */
+    private function itemIDsFromRequest(): array
+    {
+        $request = $this->getRequest();
+        $itemIDs = $request->getVar('itemID');
```
emteknetnz: Don't use getVar('itemID'), it means we have weird looking URLs such as

/admin/linkfield/data/?itemID[]=2&itemID[]=4 , instead of:
/admin/linkfield/data/2,4

Use $request->param('ItemID') instead and split on ','

src/Form/MultiLinkField.php
```diff
@@ -0,0 +1,171 @@
...
+        $attributes['data-value'] = $this->getValueArray();
+        return $attributes;
+    }
+
+    /**
+     * Extracts the value of this field, normalised as a non-associative array.
+     */
+    private function getValueArray(): array
+    {
+        return $this->cooerceValueToArray($this->Value());
```
emteknetnz:
```suggestion
        return $this->convertValueToArray($this->Value());
```

src/Form/MultiLinkField.php
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    /**
+     * Cooerces the value to an array if possible.
+     * @throws LogicException if the type cannot be cooerced into an array.
+     */
+    private function cooerceValueToArray(mixed $value): array
```
emteknetnz:
```suggestion
    private function convertValueToArray(mixed $value): array
```

src/Form/MultiLinkField.php
```diff
@@ -0,0 +1,171 @@
...
+
+        if (is_iterable($value) && !is_array($value)) {
+            return [...$value];
+        }
+
+        if (is_array($value)) {
+            return array_values($value);
+        }
+
+        throw new LogicException('Unexpected value type ' . gettype($value));
```
emteknetnz:
```suggestion
        if (is_string($value)) {
            $value = $this->convertCommaSeparatedString(trim($value));
        }
        if (empty($value)) {
            return [];
        }
        if ($value instanceof SS_List) {
            return $value->column('ID');
        }
        if (!is_iterable($value)) {
            return [$value];
        }
        if (is_iterable($value) && !is_array($value)) {
            return [...$value];
        }
        if (is_array($value)) {
            return array_values($value);
        }
        throw new LogicException('Unexpected value type ' . gettype($value));
```
All the blank lines aren't needed since there's closing brackets to visually split everything up

emteknetnz: Need to unit test this

src/Form/MultiLinkField.php
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    /**
+     * Cooerces a comma-separated string of integers into an array.
+     * If any value is not an integer, it returns the original string.
+     */
+    private function coeerceCommaSeparatedString(string $string): string|array
```
emteknetnz:
```suggestion
    private function convertCommaSeparatedString(string $string): string|array
```

client/src/components/LinkField/LinkField.js
```diff
@@ -19,36 +22,63 @@ const section = 'SilverStripe\\LinkField\\Controllers\\LinkFieldController';
...
+      linkIDs = [linkIDs];
+    }
+  }
+
+  // Read data from endpoint and update component state
+  // This happens any time a link is added or removed and triggers a re-render
+  useEffect(() => {
+    if (!editingID && linkIDs.length > 0) {
+      const query = [];
+      for (const id of linkIDs) {
```
emteknetnz:
```suggestion
      for (const linkIDs of linkIDs) {
```

client/src/components/LinkField/LinkField.js
```diff
@@ -19,36 +22,63 @@ const section = 'SilverStripe\\LinkField\\Controllers\\LinkFieldController';
...
  * actions - object of redux actions
+ * isMulti - whether this field handles multiple links or not
  */
-const LinkField = ({ value, onChange, types, actions }) => {
-  const linkID = value;
-  const [typeKey, setTypeKey] = useState('');
+const LinkField = ({ value = null, onChange, types, actions, isMulti = false  }) => {
   const [data, setData] = useState({});
-  const [editing, setEditing] = useState(false);
+  const [editingID, setEditingID] = useState(null);
```
emteknetnz:
```suggestion
  const [editingID, setEditingID] = useState(-1);
```
Are we using null because 0 is reserved for editing new unsaved Links?

`-1` is better than `null` because it's easy to make errors by testing the truthiness of editingID when 0 is a valid value. Instead explicitly test for `!== -1` similar to `indexOf()`.

client/src/components/LinkField/LinkField.js
```diff
@@ -87,69 +114,65 @@ const LinkField = ({ value, onChange, types, actions }) => {
...
-      setEditing(true);
-    },
-    types: Object.values(types)
+  /**
+   * Render all of the links currently in the field data
+   */
+  const renderLinks = () => {
+    const links = [];
+
+    for (const id of linkIDs) {
```
emteknetnz: There's a real mixtures of 'id' and 'linkID' being used interchangably.  Previously it was almost exclusively linkID, because id cannot be used as prop as it's reserved by react, so linkID was used instead.

Could you update all of the `id` to `linkID` in various places in JS

src/Controllers/LinkFieldController.php
```diff
@@ -85,20 +87,35 @@ public function linkForm(): Form
...
      * /admin/linkfield/data/<LinkID>
      */
-    public function linkData(): HTTPResponse
+    public function linkData(HTTPRequest $request): HTTPResponse
     {
-        $link = $this->linkFromRequest();
-        if (!$link->canView()) {
-            $this->jsonError(403, _t('LinkField.UNAUTHORIZED', 'Unauthorized'));
+        $data = [];
+        if ($request->param('ID')) {
```
emteknetnz: Use of $request->param('ID') is confusing as it's not in part of `private static $url_handlers` - `'linkForm/$ItemID' => 'linkForm'`, instead it's one of the default request params?  What part of the URL does 'ID' actually refer to in this case?

src/Controllers/LinkFieldController.php
```diff
@@ -274,17 +291,56 @@ private function linkFromRequest(): Link
...
+    {
+        $itemIDs = $this->itemIDsFromRequest();
+        if (empty($itemIDs)) {
+            $this->jsonError(404, _t('LinkField.INVALID_ID', 'Invalid ID'));
+        }
+        $link = Link::get()->byIDs($itemIDs);
+        if (!$link->exists()) {
+            $this->jsonError(404, _t('LinkField.INVALID_ID', 'Invalid ID'));
+        }
+        return $link;
```
```suggestion
        $links = Link::get()->byIDs($itemIDs);
        if (!$links->exists()) {
            $this->jsonError(404, _t('LinkField.INVALID_ID', 'Invalid ID'));
        }
        return $links;
```
Though I think the use of DataList::exists() isn't quite right, since that will `Returns true if this DataList has items` - need to check the number of items returns is the same as the length of $itemIDs

src/Form/MultiLinkField.php
```diff
@@ -0,0 +1,171 @@
...
+    public function setValue($value, $data = null)
+    {
+        // If $data is a record, we can pull the value directly from it.
+        // This mirrors MultiSelectField::setValue().
+        if ($data instanceof DataObject) {
+            $this->loadFrom($data);
+            return $this;
+        }
+
+        $ids = $this->cooerceValueToArray($value);
```
emteknetnz:
```suggestion
        $ids = $this->convertValueToArray($value);
```

emteknetnz (changes requested):

README.md
ChloeSartorelli: Sure. Done. Added note about adding a `has_one` to the `Link` model via yaml because that's what's required currently. I'm not going to document this to what we want to eventually have - the documentation is for the current state. When we change the way ownership works, we can update the docs for that new state.

README.md
ChloeSartorelli: Done

client/src/components/LinkField/LinkField.js
ChloeSartorelli: Done. I think the `if (!linkIDs)` condition was originally meant to go after the value check but this works just as well.

client/src/components/LinkField/LinkField.js
ChloeSartorelli: `?itemID[]=2&itemID[]=4` is not strange looking - it's the correct way to handle array query strings for PHP `GET` requests. PHP converts it into an array for us. I don't think it makes sense to pass multiple IDs as a comma-separated param. I won't be making this change as `/2,4` is not a normal way to pass these, but the array `GET` query string _is_.

client/src/components/LinkField/LinkField.js
ChloeSartorelli: We have to check the value and value length in order to trigger the effect when _new_ links are added. `editingID` is only for editing _existing_ links.

client/src/components/LinkField/LinkField.js
ChloeSartorelli: Done

client/src/components/LinkField/LinkField.js
ChloeSartorelli: Why?

client/src/components/LinkPicker/LinkPicker.js
ChloeSartorelli: Done

client/src/components/LinkPicker/LinkPicker.js
ChloeSartorelli: Done

client/src/components/LinkField/LinkField.js
ChloeSartorelli: Done to avoid ping pong but this just seems unnecessary.

client/src/components/LinkPicker/LinkPicker.js
ChloeSartorelli: Done

client/src/components/LinkPicker/LinkPicker.js
ChloeSartorelli: Done

client/src/components/LinkPicker/LinkPicker.js
ChloeSartorelli: I don't understand the reasoning but it doesn't make any material difference so I've changed it to avoid ping pong.

client/src/components/LinkPicker/LinkPicker.js
ChloeSartorelli: Can't reassign types so that's now `const typeArray`.
Done.

client/src/components/LinkPicker/LinkPicker.scss
ChloeSartorelli: `%` is a scss template - basically means "this is for use with `@extend` but won't be output in the final css on its own".
We don't need a `.link-row` in the final css.

client/src/containers/LinkModalContainer.js
ChloeSartorelli: That logic was pre-existing in linkfield, I've just carried it over to here instead. Dunno what it was doing there but I didn't want to mess with it. Probably changing how that's handled would be a separate card.

src/Controllers/LinkFieldController.php
ChloeSartorelli: Any discussion about this will be in https://github.com/silverstripe/silverstripe-linkfield/pull/120#discussion_r1409874663

src/Form/MultiLinkField.php
ChloeSartorelli: We're not converting, we're coeercing.... but I'll change this to avoid pingpong.

src/Form/MultiLinkField.php
ChloeSartorelli: > All the blank lines aren't needed since there's closing brackets to visually split everything up

We don't have any standards around that - that's entirely subjective and comes up time and again where one person wants new lines and another person doesn't. I don't think we should bother pulling each other up on these sorts of changes where we don't actually have any rules about how it should be done.

In this case I've made the change to avoid ping pong, but in the future I might push back because it really shouldn't be mentioned in the first place unless/until we have rules about it.

client/src/components/LinkField/LinkField.js
ChloeSartorelli: Done to avoid ping pong. We know what the ID is though so seems unnecessary.

client/src/components/LinkField/LinkField.js
ChloeSartorelli: We're using null because it's unset, and we can easily just do a truthy check against it. If there's an ID, it'll be truthy. I'll change to `0` if you want but `-1` is nonsensical and isn't what we use _anywhere_ to denote "not an ID".

client/src/components/LinkField/LinkField.js
ChloeSartorelli: I'm not gonna go through and look for them all because I don't think it's worth my time. We know it's a link ID. We don't need to be so explicit about it.

I've changed this instance of it to avoid ping pong.

src/Form/MultiLinkField.php
ChloeSartorelli: I've changed all of these - not bothering to comment "done" on them all.

src/Controllers/LinkFieldController.php
ChloeSartorelli: Was meant to be `ItemID` - fixed.

src/Controllers/LinkFieldController.php
ChloeSartorelli: Updated to `$links`. Also, `exists()` is correct here because we do only care if there are items in the list, not how many items there are. We explicitly passed in the array of IDs from the request. There will never be _more_ than that number of links. If there are fewer, it means that the relation was outdated and had some links in it that don't exist. Only the links that exist will be displayed, and when the user clicks save the missing link IDs will disappear.

client/src/components/LinkField/LinkField.js
ChloeSartorelli: Oops nope that actually broke the single link field. Moved the if not condition to be after the numeric check.

src/Form/MultiLinkField.php
ChloeSartorelli: Added unit test

client/src/components/LinkField/LinkField.js
emteknetnz: OK, set it to `[editingID, linkIDs.length]` since we've converted value to linkIDs

ChloeSartorelli: That results in an infinite loop, since we're manipulating the value of `linkIDs`.

client/src/components/LinkField/LinkField.js
emteknetnz: OK `0` makes sense

client/src/components/LinkField/LinkField.js
```diff
@@ -19,36 +22,63 @@ const section = 'SilverStripe\\LinkField\\Controllers\\LinkFieldController';
...
+
+  // Read data from endpoint and update component state
+  // This happens any time a link is added or removed and triggers a re-render
+  useEffect(() => {
+    if (!editingID && linkIDs.length > 0) {
+      const query = [];
+      for (const linkID of linkIDs) {
+        query.push(`itemID[]=${linkID}`);
+      }
+      const endpoint = `${Config.getSection(section).form.linkForm.dataUrl}/?${query.join('&')}`;
```
emteknetnz:
```suggestion
      const endpoint = `${Config.getSection(section).form.linkForm.dataUrl}?${query.join('&')}`;
```
Don't need the trailing slash since there's no params being used on the endpoint

Changes 
/admin/linkfield/data/?itemID[]=6&itemID[]=10&itemID[]=11&itemID[]=12
to
/admin/linkfield/data?itemID[]=6&itemID[]=10&itemID[]=11&itemID[]=12

client/src/components/LinkField/LinkField.js
```diff
@@ -19,36 +22,63 @@ const section = 'SilverStripe\\LinkField\\Controllers\\LinkFieldController';
...
+    }
+  }
+
+  // Read data from endpoint and update component state
+  // This happens any time a link is added or removed and triggers a re-render
+  useEffect(() => {
+    if (!editingID && linkIDs.length > 0) {
+      const query = [];
+      for (const linkID of linkIDs) {
+        query.push(`itemID[]=${linkID}`);
```
emteknetnz:
```suggestion
        query.push(`itemIDs[]=${linkID}`);
```

src/Controllers/LinkFieldController.php
```diff
@@ -274,17 +291,56 @@ private function linkFromRequest(): Link
...
+        return (int) $itemID;
+    }
+
+    /**
+     * Get the value of the itemID request query string argument
+     */
+    private function itemIDsFromRequest(): array
+    {
+        $request = $this->getRequest();
+        $itemIDs = $request->getVar('itemID');
```
```suggestion
        $itemIDs = $request->getVar('itemIDs');
```

emteknetnz (changes requested):
The regular singlelink field wasn't working when I tested with the latest changes, though the multilink field was working

client/src/components/LinkField/LinkField.js
ChloeSartorelli: Done

client/src/components/LinkField/LinkField.js
ChloeSartorelli: Removed to avoid ping pong, though it doesn't affect anything one way or another.

client/src/components/LinkField/LinkField.js
ChloeSartorelli: Because we don't have behat tests yet small unnecessary changes like this are more likely to break things than to be useful. I don't want to keep making these small changes that don't actually affect anything - it means I have to test every bit of functionality every time it goes through a review.

Done.

GuySartorelli:
> The regular singlelink field wasn't working when I tested with the latest changes, though the multilink field was working

It's working for me locally with this set of changes - hopefully something I did fixed whatever was broken, but if there's still something about it that isn't working please describe the actual faulty behaviour so I know what specifically needs fixing.

src/Controllers/LinkFieldController.php
ChloeSartorelli: Done

emteknetnz (approved):
Tested locally, worked good

PR #127, opened 2023-12-04
API Add new Owner relation for handling permissions

GuySartorelli:
## Dependencies
This PR requires https://github.com/silverstripe/silverstripe-framework/pull/11084 to function and for tests to go green but an installer CI run is linked in the issue.

## Issues
- https://github.com/silverstripe/silverstripe-linkfield/issues/5

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,22 +40,27 @@ class Page extends SiteTree
     ];
 
     private static $has_many = [
-        'HasManyLinks' => Link::class,
+        // Multiple has_many relations on the same class should point at the same has_one on Link.
+        'HasManyLinksOne' => Link::class . '.Owner',
+        'HasManyLinksTwo' => Link::class . '.Owner',
     ];
 
     private static array $owns = [
         'HasOneLink',
-        'HasManyLinks',
+        'HasManyLinksOne',
+        'HasManyLinksTwo',
     ];
 
     private static array $cascade_deletes = [
         'HasOneLink',
-        'HasManyLinks',
+        'HasManyLinksOne',
+        'HasManyLinksTwo',
     ];
 
     private static array $cascade_duplicates = [
         'HasOneLink',
-        'HasManyLinks',
+        'HasManyLinksOne',
+        'HasManyLinksTwo',
     ];
 
     public function getCMSFields()
@@ -63,13 +68,14 @@ class Page extends SiteTree
         $fields = parent::getCMSFields();
 
         // Don't forget to remove the auto-scaffolded fields!
-        $fields->removeByName(['HasOneLinkID', 'Links']);
+        $fields->removeByName(['HasOneLinkID', 'HasManyLinksOne', 'HasManyLinksTwo']);
 
         $fields->addFieldsToTab(
             'Root.Main',
             [
                 LinkField::create('HasOneLink'),
-                MultiLinkField::create('HasManyLinks'),
+                MultiLinkField::create('HasManyLinksOne'),
+                MultiLinkField::create('HasManyLinksTwo'),
             ],
         );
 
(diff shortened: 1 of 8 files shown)


src/Models/FileLink.php
```diff
@@ -5,6 +5,12 @@
 use SilverStripe\Assets\File;
 use SilverStripe\Forms\FieldList;
 
+/**
+ * A link to a File in the CMS
+ *
+ * @property int $FileID
+ * @method File File()
+ */
```
GuySartorelli, line 13: This isn't related but it needed to be added to avoid the PHPDoc for `Link` (which says "don't touch me" essentially) from leaking through to this class.

src/Models/Link.php
```diff
@@ -13,11 +13,15 @@
...
 use SilverStripe\LinkField\Type\Registry;
 use SilverStripe\ORM\DataObject;
+use SilverStripe\ORM\DataObjectSchema;
 use SilverStripe\ORM\FieldType\DBHTMLText;
 
 /**
- * A Link Data Object. This class should be a subclass, and you should never directly interact with a plain Link
- * instance
+ * A Link Data Object. This class should be treated as abstract. You should never directly interact with a plain Link
+ * instance.
```
GuySartorelli, line 21: Rewording this so it makes more sense.

src/Models/Link.php
```diff
@@ -31,6 +35,17 @@ class Link extends DataObject
...
+    private static array $has_one = [
+        // Note that this handles one-to-many relations AND one-to-one relations.
+        // Any has_one pointing at Link will be intentionally double handled - this allows us to use the owner
+        // for permission checks and to link back to the owner from reports, etc.
+        // See also the Owner method.
+        'Owner' => [
+            'class' => DataObject::class,
+            DataObjectSchema::HASONE_IS_MULTIRECIPROCAL => true,
+        ],
+    ];
```
ChloeSartorelli: See the associated framework PR and docs PR for more about this weird `has_one` syntax.

src/Models/Link.php
```diff
@@ -277,6 +292,78 @@ public function getURL(): string
...
+        // actually still owns this record. If not, return null.
+        $ownerRelationType = $owner->getRelationType($this->OwnerRelation);
+        if ($ownerRelationType === 'has_one') {
+            $idField = "{$this->OwnerRelation}ID";
+            if ($owner->$idField !== $this->ID) {
+                return null;
+            }
+        }
+        return $owner;
+    }
```
ChloeSartorelli: We're double handling `has_one` relations to link by storing the `has_one` _in the link as well_.

This allows us to use the `Owner` relation to check permissions, and to link to the owner in reports, etc.

The double handling is mitigated here by returning `null` if the owner says it doesn't know about the link.

README.md
```diff
@@ -40,21 +40,24 @@ class Page extends SiteTree
     ];
 
     private static $has_many = [
-        'HasManyLinks' => Link::class
+        // Multiple has_many relations on the same class should point at the same has_one on Link.
+        'HasManyLinksOne' => Link::class . '.Owner',
+        'HasManyLinksTwo' => Link::class . '.Owner',
```
emteknetnz, line 45: I know it's best practice though does anyone actually suffixed on the the relationships?  This would only be required if there were multiple polymorphoic has_one's on Link correct?

I'd prefer to keep things simpler and more 'standard' (for the average user) which is to not have the relation suffix

emteknetnz: Also means we don't need two has_many relationships for this example

src/Models/Link.php
```diff
@@ -31,6 +35,17 @@ class Link extends DataObject
...
     ];
 
+    private static array $has_one = [
+        // Note that this handles one-to-many relations AND one-to-one relations.
+        // Any has_one pointing at Link will be intentionally double handled - this allows us to use the owner
+        // for permission checks and to link back to the owner from reports, etc.
+        // See also the Owner method.
+        'Owner' => [
+            'class' => DataObject::class,
+            DataObjectSchema::HASONE_IS_MULTIRECIPROCAL => true,
```
emteknetnz: I've suggested changing this in the framework PR

emteknetnz (changes requested):

README.md
ChloeSartorelli: > Also means we don't need two has_many relationships for this example

I think it's important to document having multiple has_many relations works without creating new has_one relations, or people might default to doing what they have always had to do and add more has_one relations via yaml.

> I'd prefer to keep things simpler and more 'standard' (for the average user) which is to not have the relation suffix

Our documentation should always follow best practices. It's embarrassing if someone follows our documentation and then the relation validation service complains that they've done it wrong.

src/Models/Link.php
ChloeSartorelli: Done

emteknetnz (approved):

emteknetnz (changes requested):
Needs rebase

GuySartorelli:
Rebased and conflicts resolved

emteknetnz (approved):

Issue #103, opened 2023-10-17
LinkField honours Link permission


maxime-rainville:
As the owner of a Silverstripe CMS project, I want link data to be access and updated only by authorised content author.

## Acceptance criteria
- LinkField does not display information about link the current content author does not have CanView rights on.
- LinkField does not let you update Link data if you do not have CanEdit permission on the Link.
- LinkField does not let you delete Link data if you do not have CanDelete permission on the Link.
- LinkField does not let you create Link data if you do not have CanCreate permission on the Link.
- Any relations associated to a Link also honor access control restrictions (e.g: related page or file).
- Any APIs used to retrieve/update link data honours any relevant access control restriction.
- ~~Toast errors should still be shown when the user does a request that return an HTTP error.~~
- UI buttons to create / edit / delete links should not show if the user fails corresponding permission checks

## Excluded
- Readonly and hidden views when the user doesn't have the right update or view the link. That will come later.

## Pull request 
- https://github.com/silverstripe/silverstripe-linkfield/pull/141

GuySartorelli:
PR merged

PR #134, opened 2023-12-05
ENH Add permission methods based on owner

GuySartorelli:
## Note
The behaviour of the link field itself is out of scope - see https://github.com/silverstripe/silverstripe-linkfield/issues/103

## Issue
- https://github.com/silverstripe/silverstripe-linkfield/issues/12

### DIFF
--- a/src/Models/Link.php
+++ b/src/Models/Link.php
@@ -342,6 +342,67 @@ class Link extends DataObject
         return $owner;
     }
 
+    public function canView($member = null)
+    {
+        return $this->canPerformAction(__FUNCTION__, $member);
+    }
+
+    public function canEdit($member = null)
+    {
+        return $this->canPerformAction(__FUNCTION__, $member);
+    }
+
+    public function canDelete($member = null)
+    {
+        return $this->canPerformAction(__FUNCTION__, $member);
+    }
+
+    public function canCreate($member = null, $context = [])
+    {
+        // Allow extensions to override permission checks
+        $results = $this->extendedCan(__FUNCTION__, $member, $context);
+        if (isset($results)) {
+            return $results;
+        }
+
+        // Assume anyone can create a link by default - there's no way to determine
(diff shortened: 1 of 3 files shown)


sabina-talipova (commented):
Could you provide an example of how I could test these changes, it seems to me that I am testing it incorrectly.
Steps I took for testing:
- I have a Page
- The page has all `can*` methods that return `true`.
- I have an `ExternalLinkExtension` which has a `canView` which returns `false`.
- When I create a new link, no matter what type, I do not see the new link in the `LinkField`.
- If I change the return value of `ExternalLinkExtension::canView` to `true` then everything works.

GuySartorelli:
> - I have a Page
> - The page has all `can*` methods that return `true`.
> - I have an `ExternalLinkExtension` which has a `canView` which returns `false`.
> - When I create a new link, no matter what type, I do not see the new link in the `LinkField`.
> - If I change the return value of `ExternalLinkExtension::canView` to `true` then everything works.

Seeing the permissions be respected in the `LinkField` is out of scope - you should call the can* method directly and check the returned value is what you think it should be.
But in this case, what you're seeing is the expected behaviour.

Here's what should happen, in a variety of scenarios:

## If there are no extensions and no `Owner`
- `canView()`, `canEdit()`, and `canDelete()` should use the default value as defined in `DataObject`
- `canCreate()` should return `true`

## If there are no extensions but there is an `Owner`
- `canView()` should return the same as `$this->Owner()->canView()`
- `canEdit()` and `canDelete()` should return the same as `$this->Owner()->canEdit()`
- `canCreate()` should return `true`

## If there are extensions

Regardless of the value on the owner, or even if there is an owner or not, if the extension's can* method returns `true` or `false`, that value will be returned.

If the extension's can* method returns `null`, it is as though the extension isn't there, so use the appropriate heading above for whether there is an owner or not.

sabina-talipova (approved):
LGTM. Test locally.

PR #137, opened 2023-12-07
NEW Allowed link types

sabina-talipova:
## Description
- Added a new trait `AllowedLinkClassesTrait` that adds methods to set available link types or exclude them.
- The `Registry::class` has been removed.
- Files and methods for working with GraphQL have also been removed
- Added a service class to generate available existing Link classes

## Parent issue
- https://github.com/silverstripe/silverstripe-linkfield/issues/15


README.md
```diff
@@ -39,7 +39,7 @@ class Page extends SiteTree
         'HasOneLink' => Link::class,
     ];
 
-    private static $has_many = [
+    private static $many_many = [
```
maxime-rainville: I don't think we'll have many-many support for links.

src/Services/LinkTypeService.php
```diff
@@ -0,0 +1,101 @@
...
+use SilverStripe\LinkField\Models\Link;
+
+class LinkTypeService
+{
+    use Injectable;
+
+    /**
+     * Generate all link types that are subclasses of Link::class
+     */
+    public static function generateAllLinkTypes(): array
```
maxime-rainville: Generally, when we make an injectable service, we make it's method instance method, rather than static.

```suggestion
    public function generateAllLinkTypes(): array
```

src/Services/LinkTypeService.php
```diff
@@ -0,0 +1,101 @@
...
+        }
+
+        return $result;
+    }
+
+    /**
+     * Return a Link instance by key
+     * @throws InvalidArgumentException
+     */
+    public static function byKey(string $key): ?Link
```
maxime-rainville:
```suggestion
    public function byKey(string $key): ?Link
```

src/Services/LinkTypeService.php
```diff
@@ -0,0 +1,101 @@
...
+        }
+
+        return static::definitionToType($definition);
+    }
+
+    /**
+     * Return a key for link type by classname
+     * @throws InvalidArgumentException
+     */
+    public static function keyByClassName(string $classname): ?string
```
maxime-rainville:
```suggestion
    public function keyByClassName(string $classname): ?string
```

src/Services/LinkTypeService.php
```diff
@@ -0,0 +1,101 @@
...
+
+namespace SilverStripe\LinkField\Services;
+
+use InvalidArgumentException;
+use SilverStripe\Core\ClassInfo;
+use SilverStripe\Core\Injector\Injectable;
+use SilverStripe\Core\Injector\Injector;
+use SilverStripe\LinkField\Models\Link;
+
+class LinkTypeService
```
maxime-rainville, line 15:
```suggestion
/**
 * Class to manage which Link type can be added to LinkField form field.
 */
class LinkTypeService
```

src/Models/Link.php
```diff
@@ -188,7 +189,7 @@ function setData($data): Link
             );
         }
 
-        $type = Registry::singleton()->byKey($typeKey);
+        $type = LinkTypeService::byKey($typeKey);
```
maxime-rainville:
```suggestion
        $type = LinkTypeService::create()->byKey($typeKey);
```

src/Models/Link.php
```diff
@@ -225,7 +226,7 @@ function setData($data): Link
 
     public function jsonSerialize(): mixed
     {
-        $typeKey = Registry::singleton()->keyByClassName(static::class);
+        $typeKey = LinkTypeService::keyByClassName(static::class);
```
maxime-rainville:
```suggestion
        $typeKey = LinkTypeService::create()->keyByClassName(static::class);
```

src/Models/Link.php
```diff
@@ -320,4 +321,9 @@ public function getDefaultTitle(): string
         }
         return $default;
     }
+
+    public function getShortCode(): string
```
maxime-rainville, line 453: I would add a description.

Initially, I gave each link type a short code because I wasn't sure they would always match to an actual Link DataObject.

I think there's a simplicity argument for using the fully qualified name of Link classes instead of short code. But I'm not overly fuss about this part.

src/Services/LinkTypeService.php
```diff
@@ -0,0 +1,101 @@
...
+            }
+        }
+
+        return null;
+    }
+
+    /**
+     * @throws InvalidArgumentException
+     */
+    public static function definitionToType(string $className): Link
```
maxime-rainville:
```suggestion
    public function definitionToType(string $className): Link
```

src/Services/LinkTypeService.php
```diff
@@ -0,0 +1,101 @@
...
+            if ($class === $classname) {
+                return $key;
+            }
+        }
+
+        return null;
+    }
+
+    /**
+     * @throws InvalidArgumentException
```
maxime-rainville: Maybe add a description of the purpose of this method.

src/Services/LinkTypeService.php
```diff
@@ -0,0 +1,101 @@
...
+            throw new InvalidArgumentException(
+                _t(
+                    'LinkField.NO_CLASSNAME',
+                    '"{class}": All types should reference a valid classname',
+                    ['class' => static::class],
+                    sprintf('%s: All types should reference a valid classname', static::class),
+                ),
+            );
+        }
+
```
maxime-rainville: I would be inclined to let the Injector do its job here.

For example, I could use the Injector to map an alias for a link type. With that initial condition, that wouldn't work even if my alias maps to an actual link type via the injector.

src/Services/LinkTypeService.php
```diff
@@ -0,0 +1,101 @@
...
+        /** @var Link $type */
+        $type = Injector::inst()->get($className);
+
+        if (!$type instanceof Link) {
+            throw new InvalidArgumentException(
+                _t(
```
maxime-rainville: We don't generally translate exception we don't expect the end user to see.

src/Traits/AllowedLinkClassesTrait.php
```diff
@@ -0,0 +1,117 @@
...
+
+namespace SilverStripe\LinkField\Traits;
+
+use SilverStripe\Core\ClassInfo;
+use SilverStripe\Core\Config\Config;
+use SilverStripe\Core\Injector\Injector;
+use SilverStripe\LinkField\Models\Link;
+use SilverStripe\LinkField\Services\LinkTypeService;
+
+trait AllowedLinkClassesTrait
```
maxime-rainville: Add a description.

src/Traits/AllowedLinkClassesTrait.php
```diff
@@ -0,0 +1,117 @@
...
+    public function getTypesProps(): string
+    {
+        return json_encode($this->list());
+    }
+
+    /**
+     * @return Link[]
+     * @throws InvalidArgumentException
+     */
+    public function list(): array
```
maxime-rainville: This method looks very out of place here. Having a method called `list` is very confusing. It doesn't look like it's returning a array of Link data object either.

src/Traits/AllowedLinkClassesTrait.php
```diff
@@ -0,0 +1,117 @@
...
+
+use SilverStripe\Core\ClassInfo;
+use SilverStripe\Core\Config\Config;
+use SilverStripe\Core\Injector\Injector;
+use SilverStripe\LinkField\Models\Link;
+use SilverStripe\LinkField\Services\LinkTypeService;
+
+trait AllowedLinkClassesTrait
+{
+    protected $allowed_types = [];
```
maxime-rainville:
```suggestion
    private $allowed_types = [];
```

src/Traits/AllowedLinkClassesTrait.php
```diff
@@ -0,0 +1,117 @@
...
+
+    /**
+     * Set allowed types for LinkField
+     */
+    public function setAllowedTypes(array $types): static
+    {
+        if (!is_array($types) || empty($types) || !$this->validateTypes($types)) {
```
maxime-rainville: We don't need to explicitly test that this is an array. If you call the method with something other than an array, you'll get an exception right away.
```suggestion
        if (empty($types) || !$this->validateTypes($types)) {
```

src/Traits/AllowedLinkClassesTrait.php
```diff
@@ -0,0 +1,117 @@
...
+
+        return count($validClasses) > 0;
+    }
+
+    /**
+     * Set disabled types for LinkField
+     */
+    public function setDisabledTypes(array $types): static
+    {
+        if (is_array($types) || !empty($types)) {
```
maxime-rainville:
```suggestion
        if (!empty($types)) {
```

src/Traits/AllowedLinkClassesTrait.php
```diff
@@ -0,0 +1,117 @@
...
+            }
+        }
+
+        return count($validClasses) > 0;
+    }
+
+    /**
+     * Set disabled types for LinkField
+     */
+    public function setDisabledTypes(array $types): static
```
maxime-rainville: I think the interaction between `setAllowedTypes` and `setDisabledTypes` could be very confusing. I'm not sure what a reasonnable developer should expect when chaining those methods. My instinct would be to only have `setAllowedTypes`.

src/Traits/AllowedLinkClassesTrait.php
```diff
@@ -0,0 +1,117 @@
...
+    {
+        $typeDefinitions = !empty($types) ? $types : $this->getAllowedTypes();
+
+        if (empty($typeDefinitions)) {
+            $typeDefinitions = ClassInfo::subclassesFor(Link::class);
+        }
+
+        $result = array();
+        foreach ($typeDefinitions as $class) {
+            if (is_subclass_of($class, Link::class)) {
```
maxime-rainville: Seems to me that this should throw an exception if you give it something that doesn't exists.

src/Traits/AllowedLinkClassesTrait.php
maxime-rainville, line 1: I would be inclined to put this trait in the Form folder because it's so tightly tied to the fields.

maxime-rainville (changes requested):
That's a good solid first pass. I tested it locally and it works great.

There wasn't an explicit AC for this, but I think logically we should have an exception thrown if you try to save a link of a type that is disallowed for the field. e.g. Your LinkField is configured to only allow EmailLink, but you do a manual AJAX query to save a SiteTreeLink into it.

README.md
sabina-talipova: Link::class doesn't have relations with Page::class and if to follow this docs([See](https://docs.silverstripe.org/en/5/developer_guides/model/relations/#has-many).) user cannot use `has_many = 'Second::class'` relation in First::class without having `has_one = 'First::class'` in the Second::class.

src/Services/LinkTypeService.php
sabina-talipova: Updated

src/Services/LinkTypeService.php
sabina-talipova: Updated

src/Services/LinkTypeService.php
sabina-talipova: Updated

src/Models/Link.php

src/Models/Link.php

src/Services/LinkTypeService.php
sabina-talipova: Updated

src/Traits/AllowedLinkClassesTrait.php
sabina-talipova: Updated

src/Traits/AllowedLinkClassesTrait.php
sabina-talipova: Updated

src/Traits/AllowedLinkClassesTrait.php
sabina-talipova: Updated

src/Traits/AllowedLinkClassesTrait.php
sabina-talipova: Updated

src/Traits/AllowedLinkClassesTrait.php
sabina-talipova: Updated

src/Services/LinkTypeService.php
sabina-talipova: Updated

src/Services/LinkTypeService.php
sabina-talipova: Updated. I removed this method and moved some of the logic to the trait.

src/Traits/AllowedLinkClassesTrait.php
sabina-talipova: Updated. I moved logic to the `getTypesProps`.

src/Traits/AllowedLinkClassesTrait.php
sabina-talipova: I decided that it would be better if a complete list of all types of links was returned if the user passed something wrong or not a subclass of the Link class as a parameter.

src/Services/LinkTypeService.php
sabina-talipova: Updated. I removed this method and moved some of the logic to the trait.

src/Services/LinkTypeService.php
sabina-talipova: Updated. I removed this method and moved some of the logic to the trait.

src/Models/Link.php
sabina-talipova: Updated. This keys are really helpful on front-end side

src/Traits/AllowedLinkClassesTrait.php
sabina-talipova: Updated

README.md
maxime-rainville: That problem should have been fixed by this. https://github.com/silverstripe/silverstripe-linkfield/pull/127/

README.md
```diff
@@ -109,6 +109,41 @@ class ExternalLinkExtension extends Extension
 
 ```
 
+The user can control and specify the links allowed for creation in each link field. The `setAllowedTypes` method only includes link types that have been provided as parameters.
```
maxime-rainville:
```suggestion
## Controlling what type of links can be created in a LinkField

By default, all `Link` subclasses can be created by a LinkField. This includes any custom `Link` subclasses defined in your projects or via third party module.

Developers can control the link types allowed for individual `LinkField`. The `setAllowedTypes` method only allow link types that have been provided as parameters.
```

README.md
```diff
@@ -109,6 +109,41 @@ class ExternalLinkExtension extends Extension
...
+            'Root.Main',
+            [
+                MultiLinkField::create('HasManyLinks')
+                ->setAllowedTypes([ SiteTreeLink::class ]),
+            ],
+        );
+
+        return $fields;
+    }
+}
```
maxime-rainville: I think we can shrink this example and still have it legible.

```suggestion
$fields->addFieldsToTab(
    'Root.Main',
    [
        MultiLinkField::create('PageLinkList')
            ->setAllowedTypes([ SiteTreeLink::class ]),
        Link::create('EmailLink')
            ->setAllowedTypes([ EmailLink::class ]),
    ],
);
```

src/Form/Traits/AllowedLinkClassesTrait.php
```diff
@@ -0,0 +1,104 @@
...
+/**
+ * Trait to manage which Link type can be added to LinkField form field.
+ * This trait is used in LinkField and MultiLinkField classes.
+ */
+trait AllowedLinkClassesTrait
+{
+    private $allowed_types = [];
+
+    /**
+     * Set allowed types for LinkField
```
maxime-rainville, line 19:
```suggestion
     * Set allowed types for LinkField
     * @param string[] $types
```

src/Form/Traits/AllowedLinkClassesTrait.php
```diff
@@ -0,0 +1,104 @@
...
+    /**
+     * Get allowed types for LinkField
+     */
+    public function getAllowedTypes(): array
+    {
+        return $this->allowed_types;
+    }
+
+    /**
+     * Validate types that they are subclasses of Link
```
maxime-rainville, line 40:
```suggestion
     * Validate types that they are subclasses of Link
     * @param string[] $types
```

src/Form/Traits/AllowedLinkClassesTrait.php
```diff
@@ -0,0 +1,104 @@
...
+        $types = [];
+        $typeDefinitions = $this->genarateAllowedTypes();
+
+        foreach ($typeDefinitions as $key => $class) {
+            $types[$key] = Injector::inst()->get($class);
+        }
+
+        $typesList = [];
+
+        array_map(function (Link $type, string $key) use (&$typesList) {
```
maxime-rainville: You're not using `array_map` the way it's meant to be use here. Normally, `array_map` should have a 1-to-1 equivalence between the starting array and final array.

I think what you want to use here is `array_reduce`.

src/Form/Traits/AllowedLinkClassesTrait.php
```diff
@@ -0,0 +1,104 @@
...
+            $typesList[$key] = [
+                'key' => $key,
+                'title' => $type->LinkTypeTile(),
+                'handlerName' => $type->LinkTypeHandlerName(),
+            ];
+        }, $types, array_keys($types));
```
maxime-rainville: I think this is more concise.

```suggestion
$typesList = [];
        $typeDefinitions = $this->genarateAllowedTypes();

        foreach ($typeDefinitions as $key => $class) {
            /** Link $type */
            $type = Injector::inst()->get($class);
            $typesList[$key] = [
                'key' => $key,
                'title' => $type->LinkTypeTile(),
                'handlerName' => $type->LinkTypeHandlerName(),
            ];
        }
```

src/Models/Link.php
```diff
@@ -10,7 +10,8 @@
 use SilverStripe\Forms\DropdownField;
 use SilverStripe\Forms\FieldList;
 use SilverStripe\Forms\RequiredFields;
-use SilverStripe\LinkField\Type\Registry;
+use SilverStripe\LinkField\Form\LinkField;
```
maxime-rainville: This import is not used.
```suggestion
```

src/Models/Link.php
maxime-rainville: Happy to use the shortcode if we think there's value in it. A couple things thought.

## LinkTypeTile
`LinkTypeTile` is probably meant to be spelled `LinkTypeTitle`. This method is an artefact from when we had a LinkType interface and Link Types didn't need to be DataObject.

I would suggest deleting this method and using `i18n_singular_name` wherever we would have used `LinkTypeTile`

## ClassInfo::shortName()
We should use `ClassInfo::shortName($this)` for the short code instead of the user friendly name. I don't think it matters if we live the `Link` bit in there since the short code won't be displayed to the user in the end.

tests/php/Traits/AllowedLinkClassesTraitTest.php
```diff
@@ -0,0 +1,86 @@
...
+     * @dataProvider allowedTypesDataProvider
+     */
+    public function testSetAllowedTypes(array $enabled, array $expected)
+    {
+        $trait = LinkField::create('Test');
+        $trait->setAllowedTypes($enabled);
+        $this->assertEquals($expected, $trait->getAllowedTypes());
+    }
+
+    public function allowedTypesDataProvider() : array
```
maxime-rainville, line 49: Some of those test case will fail if the project has a custom link type.

If I define a `FooLink` that extend `Link`, it will be returned in the list of allowed Link types and will cause this unit test to break.

src/Form/Traits/AllowedLinkClassesTrait.php
```diff
@@ -0,0 +1,104 @@
...
+        return count($validClasses) > 0;
+    }
+
+    /**
+     * The method returns an associational array converted to a JSON string,
+     * of available link types with additional parameters necessary
+     * for full-fledged work on the client side.
+     * @throws InvalidArgumentException
+     */
+    public function getTypesProps(): string
```
maxime-rainville, line 85: The field doesn't currently work in Elemental.

When you load the LinkField in an Elemental block, the field gets bootstrap in a  different way than with Entwine.

The props have to be passed added to the return array of `getSchemaData`

You can look at this class for a good example: https://github.com/maxime-rainville/silverstripe-react/blob/0.1/src/ReactFormField.php

src/Models/Link.php
maxime-rainville: This method is not covered by unit test.

src/Services/LinkTypeService.php
```diff
@@ -0,0 +1,69 @@
...
+use SilverStripe\Core\ClassInfo;
+use SilverStripe\Core\Injector\Injectable;
+use SilverStripe\Core\Injector\Injector;
+use SilverStripe\LinkField\Models\Link;
+
+/**
+ * This service class helps to obtain a list of all available Link subclasses,
+ * an instance of the Link subclass by key, or a key by class name.
+ */
+class LinkTypeService
```
maxime-rainville, line 15: There's no unit test for this service.

src/Traits/AllowedLinkClassesTrait.php
maxime-rainville: If you're passing something that doesn't map to an actual Link DataObject, there's a high likelihood that there's a mistake in your code.

Logically, we should throw an exception in that scenario.

maxime-rainville (changes requested):
My biggest concern is that this won't work with Elemental.

I left a link to how to react props so it works with Elemental. If it doesn't make sense let me know, and we can go through it together.

README.md
sabina-talipova: Updated

README.md
sabina-talipova: Updated

src/Form/Traits/AllowedLinkClassesTrait.php
sabina-talipova: Updated

src/Form/Traits/AllowedLinkClassesTrait.php
sabina-talipova: Updated

src/Form/Traits/AllowedLinkClassesTrait.php

src/Form/Traits/AllowedLinkClassesTrait.php

src/Models/Link.php

src/Models/Link.php
sabina-talipova: Updated

src/Services/LinkTypeService.php
sabina-talipova: Updated

src/Form/Traits/AllowedLinkClassesTrait.php
sabina-talipova: Updated. 
@maxime-rainville, could you test this with `private static bool $inline_editable = true;`

tests/php/Traits/AllowedLinkClassesTraitTest.php
sabina-talipova: Updated

sabina-talipova:
https://github.com/silverstripe/silverstripe-linkfield/pull/137#discussion_r1424807464
Updated. Now user gets `InvalidArgumentException` if he's trying to pass empty array or non-Link type classes in `setAllowedTypes`.

src/Models/Link.php
sabina-talipova: Replaced `LinkTypeTile` with `ClassInfo::shortName`.

src/Models/Link.php
```diff
@@ -446,4 +441,17 @@ public function getDefaultTitle(): string
...
+     * This method process the defined singular_name of Link class
+     * to get the short code of the Link class name.
+     * Or If the name is not defined (by redefining $singular_name in the subclass),
+     * this use the class name. The Link prefix is removed from the class name
+     * and the resulting name is converted to lowercase.
+     * Example: Link => link, EmailLink => email, FileLink => file, SiteTreeLink => sitetree
+     */
+    public function getShortCode(): string
+    {
+        return strtolower(str_replace([' ', 'Link'], '', ClassInfo::shortName($this))) ?? '';
```
maxime-rainville: This is a minor tweak.
- `ClassInfo::shortName` won't return spaces, so you don't need to remove spaces.
- It would a bit safer to only remove the `Link` part if it's at the end of the class name. So using `rtrim` would be a bit safer here.



```suggestion
        return strtolower(rtrim(ClassInfo::shortName($this), 'Link')) ?? '';
```

src/Form/Traits/AllowedLinkClassesTrait.php
maxime-rainville: Works great and both usecase.

src/Form/Traits/AllowedLinkClassesTrait.php
```diff
@@ -0,0 +1,124 @@
...
+ */
+trait AllowedLinkClassesTrait
+{
+    private array $allowed_types = [];
+
+    /**
+     * Set allowed types for LinkField
+     * @param string[] $types
+     */
+    public function setAllowedTypes(array $types): static
```
maxime-rainville, line 22: What's our thinking when an empty array is passed to this method?
- Do we want to throw an exception?
- Do we want to unset any restriction and go back to allowing all Link types?

templates/SilverStripe/LinkField/Form/LinkField.ss
```diff
@@ -1,2 +1,2 @@
 <input $AttributesHTML />
-<div data-field-id="$ID" data-schema-component="$SchemaComponent" class="entwine-linkfield"></div>
+<div data-field-id="$ID" data-schema-component="$SchemaComponent" class="entwine-linkfield" data-types="$TypesProps"></div>
```
maxime-rainville, line 2: That doesn't need to be handle here and now, but we'll probably want to pass all our props in a single attribute here. Otherwise, every time we'll need to make our bootstrapping logic more complex to handle things like read only state.

That's more of a comment. Nothing to action at this moment.

tests/php/Services/LinkTypeServiceTest.php
```diff
@@ -0,0 +1,116 @@
...
+                  'phone' => PhoneLink::class,
+                  'sitetree' => SiteTreeLink::class,
+                  'testphone' => TestPhoneLink::class,
+                ],
+            ],
+        ];
+    }
+
+    /**
+     * @dataProvider allLinkTypesDataProvider
```
maxime-rainville: If you have only one test case, you shouldn't be using a data provider.

maxime-rainville (changes requested):
It's almost there. Just a few ambiguities left to address.

src/Form/Traits/AllowedLinkClassesTrait.php
sabina-talipova: I put my comment here https://github.com/silverstripe/silverstripe-linkfield/pull/137#issuecomment-1859327356.
I think an empty array is invalid argument.

src/Models/Link.php

tests/php/Services/LinkTypeServiceTest.php
sabina-talipova: Updated

maxime-rainville (approved):
Happiness has been achieved.

Issue #140, opened 2023-12-13
Use proper namespace for PHP translation


maxime-rainville:
We made the module localisable in https://github.com/silverstripe/silverstripe-linkfield/issues/113

However we don't appear to be using our regular convention for PHP translation key.

e.g.:

This line

https://github.com/silverstripe/silverstripe-linkfield/blob/7f65d4fdee873fc7746acdc82e3bc2b591138a02/src/Models/Link.php#L106

should be written

```php
_t(__CLASS__ . '.LINK_TYPE_TITLE', 'Link Type'),
// OR
_t(Link::class . '.LINK_TYPE_TITLE', 'Link Type'),
```

## Acceptance criteria
- All PHP translation keys use the standard approach used in other module.

## PR
- https://github.com/silverstripe/silverstripe-linkfield/pull/164

GuySartorelli:
@emteknetnz Can you please link to any relevant PRs in the description?

GuySartorelli:
PR merged

Issue #11, opened 2020-11-23
Read only


maxime-rainville:
## Story
As a **CMS User** I want a **to know when a LinkField is in a readonly state** so that **I have a clear understanding that I can not edit this link**.

## Acceptance criteria
- [x] A read only UI for the link field has been designed and implemented in consultation with design team.
- [x] User can still look at the link data in the modal, but cannot interact with it when the field is in a Read only state.
- [x] Consider what to do with the existing `$this->getOwner()->canEdit()` checks on `LinkField.php` and `MultiLinkField.php` - see https://github.com/silverstripe/silverstripe-linkfield/pull/141#discussion_r1433406199

## Exclude
- Handling list of links with mix-permission.

## Note
- The Modal already has canEdit checks, so part is partially working.
- There's already checks at the API level to make sure you can't write object you don't have access to.

## PR
- https://github.com/silverstripe/silverstripe-linkfield/pull/172
- https://github.com/silverstripe/silverstripe-admin/pull/1652

maxime-rainville:
@silverstripeux ^

sachajudd:
Please see updated [designs](https://projects.invisionapp.com/dsm/silver-stripe/silverstripe-cms/nav/build/folder/components/5fbc28d0096742ec25194c31?mode=preview).

sabina-talipova:
Design team asked about the following changes:
- 'Update' button be grey instead of green so that is looks disabled.
- Fields to not be clickable at all. There should also be some sort of note in the pop-up saying that they don't have permission to edit the link.

sabina-talipova:
Design team asked about info message for read only form. Something like this.

![Screenshot 2024-01-19 at 11 54 13 AM](https://github.com/silverstripe/silverstripe-linkfield/assets/87288324/ce09d18d-7fb4-466b-9914-87c7b843bb51)

emteknetnz:
@sabina-talipova I've had a look at the AC `Consider what to do with the existing $this->getOwner()->canEdit() checks on LinkField.php and MultiLinkField.php - see https://github.com/silverstripe/silverstripe-linkfield/pull/141#discussion_r1433406199` and it seems like we don't need to change anything

I've ticked of the checkbox for that AC

sabina-talipova:
[...]

PR #141, opened 2023-12-18
ENH Tidy up permissions

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-linkfield/issues/103

This PR does a few things:
- Checks canView() for related data on SiteTreeLink and FileLink
- Add unit tests for existing logic to make the link form readonly if cannot canEdit() or canCreate()
- Makes the react 'clear' (delete) button display conditional on Link.canDelete()
- Makes the react 'create' "button" or "cannot create" div display conditional on Link.canCreate()
- Uncomments a bunch of mistakenly commented out dataProvider items which all passed when uncommented


tests/php/Controllers/LinkFieldControllerTest.php
```diff
@@ -230,15 +231,15 @@ public function provideLinkFormPost(): array
         // note: not duplicating code paths already tested with provideLinkFormGetSchema()
         // e.g. Reject Invalid ID
         return [
-            // 'Valid update existing record' => [
```
emteknetnz, line 223: A lot of these items in the data provider were previously merged while commented out, I've uncommented them

babel.config.json
```diff
@@ -0,0 +1,6 @@
+{
+  "presets": [
+    "@babel/preset-env",
+    "@babel/preset-react"
+  ]
+}
```
ChloeSartorelli, line 6: This isn't needed and was intentionally removed in https://github.com/silverstripe/silverstripe-linkfield/pull/120 - these two presets are already defined in the global webpack config.
See https://github.com/silverstripe/webpack-config/blob/5832637ef465be20afd50407cdae0185880166b5/js/modules.js#L15-L25

client/src/components/LinkPicker/LinkPicker.js
```diff
@@ -41,9 +41,17 @@ const LinkPicker = ({ types, onModalSuccess, onModalClosed }) => {
...
+      <div className={className}>
+        <div className="link-picker__cannot-create">Cannot create link</div>
+      </div>
+    );
+  }
+
   return (
     <div className={className}>
-      <LinkPickerMenu types={typeArray} onSelect={handleSelect} />
+      <LinkPickerMenu types={typeArray} onSelect={handleSelect} canCreate={canCreate} />
```
ChloeSartorelli: We're not even rendering the menu if we can't create, so we don't need to pass through this prop - the menu can just assume that if it's rendered it's allowed to create.

client/src/components/LinkPicker/LinkPicker.js
```diff
@@ -41,9 +41,17 @@ const LinkPicker = ({ types, onModalSuccess, onModalClosed }) => {
   const className = classnames('link-picker', 'form-control');
   const typeArray = Object.values(types);
 
+  if (!canCreate) {
+    return (
+      <div className={className}>
+        <div className="link-picker__cannot-create">Cannot create link</div>
```
ChloeSartorelli: Make this text translatable.

GuySartorelli (changes requested):
## canCreate

With `canCreate` returning false, it still looks like I can add links: 
image

The modal is set to readonly, but I shouldn't even be able to get to the modal.

## canView for file and sitetree

It feels a bit weird that the "File Link:" and "SiteTree Link:" are just left blank - maybe we should have a "Cannot view file/page" or similar message there in red instead?

Also, do we have opinions about how the modal should treat those?
i.e:
- the treedropdown is still editable and shows the name of the page
    - Should it be editable if the user doesn't have view access to the page currently stored there?
    - Should the page title be visible in that field?
- The file upload field is still editable, but shows nothing (as though no file was saved)
    - Should it be editable?
    - Should it provide some feedback that there _is_ a file saved there, but the current user just isn't allowed to see it?

babel.config.json
emteknetnz: Breaks jest if this is removed, when we run `yarn test` we get

`SyntaxError: /path/to/vendor/silverstripe/linkfield/client/src/components/LinkPicker/tests/LinkPickerTitle-test.js: Support for the experimental syntax 'jsx' isn't currently enabled (21:32):`

client/src/components/LinkPicker/LinkPicker.js
emteknetnz: Updated

client/src/components/LinkPicker/LinkPicker.js
emteknetnz: Removed and removed from LinkPickerMenu

emteknetnz:
- Have added a trait to LinkField/MultiLinkField that gets the Owner object from the Form, this allows us to check canCreate() on the Owner rather than using $field->Readonly()
- Have added a 'Cannot view file/page' styles messages instead of empty string
- Re the 'editable' bits for treedropdown and file upload - This PR adds a 'Cannot create link' if it fails a canCreate() check - this means you can't open the modal

babel.config.json
ChloeSartorelli: Oh, weird! I wonder if that's why I added it in the various modules when I did the CMS 5 upgrade...
Too bad we can't add comments to json files.

src/Form/MultiLinkField.php
```diff
@@ -72,13 +74,15 @@ public function getSchemaStateDefaults()
...
         $data['value'] = $this->getValueArray();
+        $data['canCreate'] = $this->getOwner()->canCreate();
         return $data;
     }
 
     protected function getDefaultAttributes(): array
     {
         $attributes = parent::getDefaultAttributes();
         $attributes['data-value'] = $this->getValueArray();
+        $attributes['data-can-create'] = $this->getOwner()->canCreate();
```
ChloeSartorelli: We don't care if we can create a page - we care if we can create a link, which is:
```php
$this->getOwner()->canEdit() && Link::singleton()->canCreate()
```

We don't need to know if we can _create_ the owner. We just need to know if we can edit the owner, or create a link. If EITHER of those is false, we can't create the link. But creating the owner is irrelevant - the owner is _already_ created.

ChloeSartorelli: Please also update this anywhere else you're doing a canCreate for links (e.g.`getSchemaStateDefaults()` above and in `LinkField`).
Might make sense to add it to the trait.

src/Models/FileLink.php
```diff
@@ -30,7 +30,14 @@ public function getCMSFields(): FieldList
 
     public function getDescription(): string
     {
-        return $this->File()?->getFilename() ?? '';
+        $file = $this->File();
+        if (!$file?->exists()) {
+            return _t('LinkField.FILE_DOES_NOT_EXIST', 'File does not exist');
```
ChloeSartorelli:
```suggestion
            return _t(__CLASS__ . '.FILE_DOES_NOT_EXIST', 'File does not exist');
```
Same with any other PHP translation strings you're adding.
Some of the pre-existing ones are not set up like this, but they will be. See https://github.com/silverstripe/silverstripe-linkfield/issues/140

GuySartorelli (changes requested):
If I set `Link::canCreate()` to return false, I still get the option to open the modal. The changes I've requested in the code should fix this.

> Re the 'editable' bits for treedropdown and file upload - This PR adds a 'Cannot create link' if it fails a canCreate() check - this means you can't open the modal

This isn't about whether I can create a link. The scenarios is this:
- I already have a link to a file or to a page
- I do not have `canView()` permissions for the file/page that is being _linked to_
- I click on the link to edit it (which I _can do_ - I _do_ have edit permissions for the link itself)
  - The modal is fully editable. See https://github.com/silverstripe/silverstripe-linkfield/pull/141#pullrequestreview-1788110436 for how that modal behaves.

GuySartorelli:
I haven't tested this in a gridfield record or in an elemental block yet - just testing on a page until everything is working correctly in that context, then I'll check other contexts.

src/Models/FileLink.php
emteknetnz: Updated

src/Form/LinkField.php
```diff
@@ -60,10 +62,18 @@ public function saveInto(DataObjectInterface $record)
         return $this;
     }
 
+    public function getSchemaStateDefaults()
+    {
+        $data = parent::getSchemaStateDefaults();
+        $data['canCreate'] = $this->getOwner()->canEdit() && Link::singleton()->canCreate();
```
ChloeSartorelli:
```suggestion
        $data['canCreate'] = $this->getOwner()->canEdit();
```
You made a good point while we were discussing this in person - we have to respect the actual type of link being created - if we can create a `PhoneLink` but not `Link` then we need to respect that. You're doing that with the change to `AllowedLinkClassesTrait` so we should remove the `canCreate()` check from here.

You can also remove the `canEdit` (and this entire prop and its functionality) if you want - if owner's `canEdit()` method returns `false` then  the field will be set to readonly (which is being handled separately in https://github.com/silverstripe/silverstripe-linkfield/issues/11) so this prop won't matter in that scenario anyway.

We should probably also add some logic (assuming it's not there already) so that when there are _no_ available link types for the field, it has a nice message instead of a weirdly empty link type menu - though I'd be okay with that being spun off separately.

ChloeSartorelli: Please also change this in the other places that pass this prop.

GuySartorelli (changes requested):
Nearly there.

I've spun off my UX inconsistency concerns into a separate card: https://github.com/silverstripe/silverstripe-linkfield/issues/147

src/Form/LinkField.php
emteknetnz: Updated

Have added AC to the [readonly card](https://github.com/silverstripe/silverstripe-linkfield/issues/11) to revisit the canEdit() logic, decide what to do with it then

Have created [new card](https://github.com/silverstripe/.github/issues/177) to handle the empty list scenario

GuySartorelli (approved):
LGTM, works well locally

Issue #145, opened 2023-12-19
Save relations on Link creation instead of parent save


emteknetnz:
Currently the has_one relations between Link and parent for both the single and the multi link field are updated when the parent is saved e.g. the page edit form is saved

This can result in a fairly uninuitive situation where orphaned links can be created if the use navigates away from the page without saving.  This will get particularly confusing once [link sorting](https://github.com/silverstripe/silverstripe-linkfield/issues/139) has been implemented

## Excluded
- Updating the parent save/published button state - this will be done in [another card](https://github.com/silverstripe/silverstripe-linkfield/issues/152)

## Acceptance criteria
- [ ] Link relations are saved to the database on Link creation (modal save) rather than parent save

## PRs
- https://github.com/silverstripe/silverstripe-linkfield/pull/146

GuySartorelli:
Merged PR

PR #146, opened 2023-12-20
ENH Save relations on link creation

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-linkfield/issues/145


GuySartorelli (changes requested):
Can you please rebase? The diff is a little weird at the moment, looks like tests are being removed but I don't think they are.

emteknetnz:
Done

client/src/components/LinkField/LinkField.js
```diff
@@ -155,19 +178,27 @@ const LinkField = ({ value = null, onChange, types = [], actions, isMulti = fals
...
     { renderModal && <LinkModalContainer
         types={types}
         typeKey={data[editingID]?.typeKey}
         isOpen={Boolean(editingID)}
         onSuccess={onModalSuccess}
         onClosed={onModalClosed}
         linkID={editingID}
+        ownerID={ownerID}
+        ownerClass={ownerClass}
+        ownerRelation={ownerRelation}
```
```suggestion
```
No need for props since we have the context.

src/Controllers/LinkFieldController.php
```diff
@@ -135,6 +135,15 @@ public function linkDelete(): HTTPResponse
...
         // delete() will also delete any published version immediately
         $link->delete();
+        // Update owner object if this Link is on a has_one relation on the owner
+        $owner = $this->ownerFromRequest();
+        $ownerRelation = $this->ownerRelationFromRequest();
+        $hasOne = Injector::inst()->get($owner->ClassName)->hasOne();
+        if (array_key_exists($ownerRelation, $hasOne) && $owner->canEdit()) {
+            $owner->$ownerRelation = null;
+            $owner->write();
+        }
```
ChloeSartorelli, line 145: I wonder if this should be in the `Link` model's `onAfterDelete()` instead, so it just happens any time a link is deleted, regardless of how it's deleted?
Not fussed either way but thought I'd raise it in case you think that's a better approach.

src/Controllers/LinkFieldController.php
```diff
@@ -215,6 +224,15 @@ public function save(array $data, Form $form): HTTPResponse
...
         }
 
+        // Update owner object if this Link is on a has_one relation on the owner
+        $owner = $this->ownerFromRequest();
+        $ownerRelation = $this->ownerRelationFromRequest();
+        $hasOne = Injector::inst()->get($owner->ClassName)->hasOne();
+        if (array_key_exists($ownerRelation, $hasOne) && $owner->canEdit()) {
+            $owner->$ownerRelation = $link;
+            $owner->write();
+        }
```
ChloeSartorelli: Do we need to do this for every edit? Seems like it should just be on create.

ChloeSartorelli: Probably also add a comment "not needed for has_many because that's stored directly on the link"
I know that's sort-of self-explanatory but I spend a few seconds going "wait why is has_many not being dealt with here?"

src/Controllers/LinkFieldController.php
```diff
@@ -215,6 +224,15 @@ public function save(array $data, Form $form): HTTPResponse
             $link->write();
         }
 
+        // Update owner object if this Link is on a has_one relation on the owner
+        $owner = $this->ownerFromRequest();
+        $ownerRelation = $this->ownerRelationFromRequest();
```
ChloeSartorelli, line 232: Do we really need to check the owner from request at this stage? The link knows who its owner is already (`$link->Owner()`)

src/Form/MultiLinkField.php
```diff
@@ -41,33 +37,15 @@ public function setValue($value, $data = null)
...
-
-        // Use RelationList rather than Relation here since some Relation classes don't allow setting value - but RelationList does.
-        if (!($relation instanceof RelationList || $relation instanceof UnsavedRelationList)) {
-            throw new LogicException("'$fieldName()' method on {$record->ClassName} doesn't return a relation list");
-        } else {
-            $relation->setByIDList($this->getValueArray() ?? []);
-        }
-
-        return $this;
-    }
```
ChloeSartorelli, line 64: We're deleting this but keeping the one in `LinkField` without any changes? What's the reasoning there?

tests/php/Form/LinkFieldTest.php
ChloeSartorelli, line 1: Why is this being deleted? That method is still there.

src/Form/Traits/LinkFieldGetOwnerTrait.php
```diff
@@ -18,4 +18,19 @@ private function getOwner(): DataObject
...
     }
+
+    private function getOwnerFields(): array
+    {
+        $owner = $this->getOwner();
+        $relation = $this->getName();
+        // Elemental content block
+        if (preg_match('#^PageElements_[0-9]+_(.+)$#', $relation, $matches)) {
+            $relation = $matches[1];
+        }
```
ChloeSartorelli: Use `ElementalAreaController::removeNamespacesFromFields()` instead.
```suggestion
        if ($owner instanceof  BaseElement) {
            $someVariableName = ElementalAreaController::removeNamespacesFromFields([$relation => ''], $owner->ID);
            $relation = array_keys($someVariableName)[0];
        }
```
I haven't tested that code but something like that should work.

ChloeSartorelli: We probably also need a unit test for this

GuySartorelli (changes requested):
This PR has changed the order in which allowed links classes are displayed. The old order was better as it seems to be from top to bottom the most to least likely type of link to be added.
Before:
image
After:
image

Other than that, *mostly* just questions rather than actually requesting changes this time around.

GuySartorelli:
I get "Invalid Owner" when I'm trying this in a record which _does already exist_ in a gridfield.
Happens for has_many _and_ has_one and for all link types.

GuySartorelli:
Obviously doesn't work when trying to add a link to a record that hasn't been saved to the DB yet - but we'll handle that in a separate card, it's not in scope for this.

emteknetnz:
Link order changed in this PR - https://github.com/silverstripe/silverstripe-linkfield/pull/137

client/src/components/LinkField/LinkField.js
emteknetnz: Removed

tests/php/Form/LinkFieldTest.php
emteknetnz: Have restroed

emteknetnz: Method shouldn't be there, have deleted

src/Form/MultiLinkField.php
emteknetnz: Have deleted LinkField::saveInto()

src/Controllers/LinkFieldController.php
emteknetnz: I think I prefer it in the controller, we're using that to related the two objects, seems sensible to also unrelated the objects in the controller too

Also this in only for has_one relations, if we put in in Link::onAfterDelete() it'll try and do it for has_many's as well which it shouldn't

src/Controllers/LinkFieldController.php
emteknetnz: Have added an line comment
```
        // Get owner using ownerFromRequest() rather than $link->Owner() so that validation is run
        // on the owner params before updating the database
```

src/Controllers/LinkFieldController.php
emteknetnz: Have added a check to see that `$owner->"{$ownerRelation}ID" === $link->ID`, meaning it will only do this if there's something to change.  This doesn't strictly limit it to on create (would use `0` instead of `$link->ID`), however there's some outside risk that somehow a related Link gets deleted and then you'll never be able to update the relation on the Owner ever again

Have added inline comment `// Only do this for has_one, not has_many, because that's stored directly on the Link record`

src/Form/Traits/LinkFieldGetOwnerTrait.php
emteknetnz: Done

src/Controllers/LinkFieldController.php
ChloeSartorelli: We have `$operation = 'create';` to check if the link is being created - so we should probably explicitly check that and _only_ do this if `$operation === 'create'`

GuySartorelli (changes requested):
Just https://github.com/silverstripe/silverstripe-linkfield/pull/146#discussion_r1434609859 left

src/Controllers/LinkFieldController.php

GuySartorelli (approved):
LGTM

Issue #150, opened 2023-12-22
Bump framework dependency to ^5.2


emteknetnz:
Because of the has_one reciprocal relationship work done that will be released framework 5.2, we need to bump the dependency of of silverstripe/framework on linkfield to ^5.2

## PR
- https://github.com/silverstripe/silverstripe-linkfield/pull/154

GuySartorelli:
PR merged

PR #154, opened 2024-01-08
DEP Explicitly require framework ^5.2

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-linkfield/issues/150

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -4,6 +4,7 @@
     "type": "silverstripe-vendormodule",
     "require": {
         "php": "^8.1",
+        "silverstripe/framework": "^5.2",
         "silverstripe/cms": "^5",
         "silverstripe/versioned": "^2"
     },


composer.json
```diff
@@ -4,8 +4,9 @@
     "type": "silverstripe-vendormodule",
     "require": {
         "php": "^8.1",
-        "silverstripe/cms": "^5",
-        "silverstripe/versioned": "^2"
+        "silverstripe/framework": "^5.2",
+        "silverstripe/cms": "^5.2",
+        "silverstripe/versioned": "^2.2"
```
ChloeSartorelli:
```suggestion
        "silverstripe/cms": "^5",
        "silverstripe/versioned": "^2"
```
No need to bump these.

GuySartorelli (changes requested):

composer.json
emteknetnz: Updated

GuySartorelli (approved):
LGTM

Issue #149, opened 2023-12-21
Change "clear" button to "delete"


GuySartorelli:
Currently there is a "clear" button next to links in the link field which deletes the link.

The text "clear" could be interpreted as "remove the link from this relation, but do not delete the link". We should swap this to be the unambiguous "delete" instead.

## PR
- https://github.com/silverstripe/silverstripe-linkfield/pull/155

GuySartorelli:
PR merged

PR #155, opened 2024-01-08
ENH Change clear text to delete

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-linkfield/issues/149

Even though the ACs are to just change the text label, it also makes sense to change all references in the codebase from 'clear' to 'delete' so that we're not using two different word for the same thing


client/lang/en.js
```diff
@@ -10,6 +10,6 @@ if (typeof(ss) === 'undefined' || typeof(ss.i18n) === 'undefined') {
     "LinkField.CONFIRM_DELETE": "Deleted link",
     "LinkField.DELETE_ERROR": "Failed to delete link",
     "LinkField.ADD_LINK": "Add Link",
-    "LinkField.CLEAR": "Clear"
+    "LinkField.Delete": "Delete"
```
ChloeSartorelli:
```suggestion
    "LinkField.DELETE": "Delete"
```
I would normally ignore this, except that it's being referenced with all caps in the code so needs to be all caps here.

GuySartorelli (changes requested):
Works well locally. Just one small change.

GuySartorelli:
I just had a thought - and feel free to say it's out of scope or not worth doing. But should this say "Archive" when links are versioned, and "Delete" when they're not? That _is_ more accurate, but I'm on the fence as to whether it's valuable enough of a distinction to be worth adding the complexity.
(Note the functionality is already doing that, so it's just whether we dynamically change the wording or not)

emteknetnz:
Have added in archived text functionality

GuySartorelli (approved):
Great! Works locally.

Issue #9, opened 2020-11-23
Used On table


maxime-rainville:
This should already work automatically, though we need to double check it works

## Acceptance criteria
- There is a relationship between FileLink and Link object in the asset-admin used of table when you click on file details

## PRs
- https://github.com/silverstripe/silverstripe-framework/pull/11107
- https://github.com/silverstripe/silverstripe-linkfield/pull/157

sachajudd:
@silverstripeux

GuySartorelli:
PRs merged

PR #157, opened 2024-01-09
ENH Used on table extension

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-linkfield/issues/9

Used on table when LinkField in on a page

image

Used on table when LinkField in on a content block on a page

image

### DIFF
--- a/src/Extensions/UsedOnTableExtension.php
+++ b/src/Extensions/UsedOnTableExtension.php
@@ -1,0 +1,32 @@
+<?php
+
+namespace SilverStripe\LinkField\Extensions;
+
+use DNADesign\Elemental\Models\BaseElement;
+use DNADesign\Elemental\TopPage\DataExtension as TopPageExtension;
+use SilverStripe\Core\Extension;
+use SilverStripe\LinkField\Models\FileLink;
+use SilverStripe\ORM\DataObject;
+
+class UsedOnTableExtension extends Extension
+{
+    public function updateUsageAncestorDataObjects(array &$ancestorDataObjects, DataObject $dataObject): void
+    {
+        if (!is_a($dataObject, FileLink::class)) {
+            return;
+        }
+        $owner = $dataObject->Owner();
+        if (!$owner?->exists()) {
+            return;
+        }
+        $ancestorDataObjects[] = $owner;
+        if (!class_exists(BaseElement::class) || !is_a($owner, BaseElement::class)) {
+            return;
+        }
+        $page = $owner->hasExtension(TopPageExtension::class) ? $owner->getTopPage() : $owner->getPage();
+        if (!$page?->exists()) {
+            return;
+        }
+        $ancestorDataObjects[] = $page;
+    }
+}
--- a/_config/config.yml
+++ b/_config/config.yml
@@ -4,3 +4,6 @@ Name: linkfield
 SilverStripe\Admin\LeftAndMain:
   extensions:
     - SilverStripe\LinkField\Extensions\LeftAndMainExtension
+SilverStripe\Admin\Forms\UsedOnTable:
+  extensions:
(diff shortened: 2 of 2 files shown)


src/Extensions/UsedOnTableExtension.php
```diff
@@ -0,0 +1,29 @@
...
+use SilverStripe\ORM\DataObject;
+
+class UsedOnTableExtension extends Extension
+{
+    public function updateUsageAncestorDataObjects(array &$ancestorDataObjects, DataObject $dataObject): void
+    {
+        if (!is_a($dataObject, FileLink::class)) {
+            return;
+        }
+        $owner = $dataObject->Owner();
```
ChloeSartorelli, line 18:
```suggestion
        $owner = $dataObject->Owner();
        if (!$owner || !$owner->exists()) {
            return;
        }
```
Unless I'm misunderstanding, we probably don't want to add a null or a dummy record into the ancestor array.

Note that while the `exists()` call is currently redundant in this context, it's good practice and protects against possible changes in the future.

src/Extensions/UsedOnTableExtension.php
```diff
@@ -0,0 +1,29 @@
...
+        $ancestorDataObjects[] = $owner;
+        if (class_exists(BaseElement::class)
+            && is_a($owner, BaseElement::class)
+            && $owner->hasMethod('getTopPage')
+        ) {
+            $page = $owner->getTopPage();
+            if ($page) {
```
ChloeSartorelli:
```suggestion
            if ($page && $page->exists()) {
```
Note that while the `exists()` call is currently redundant in this context, it's good practice and protects against possible changes in the future.

src/Extensions/UsedOnTableExtension.php
```diff
@@ -0,0 +1,29 @@
...
+        if (!is_a($dataObject, FileLink::class)) {
+            return;
+        }
+        $owner = $dataObject->Owner();
+        $ancestorDataObjects[] = $owner;
+        if (class_exists(BaseElement::class)
+            && is_a($owner, BaseElement::class)
+            && $owner->hasMethod('getTopPage')
+        ) {
+            $page = $owner->getTopPage();
```
ChloeSartorelli:
```suggestion
        ) {
            $page = $owner->hasExtension(\DNADesign\Elemental\TopPage\DataExtension) ? $owner->getTopPage() : $owner->getPage();
```
Or just use `getPage()` directly.

Also obviously use a use statement for the extension FQCN - probably with an alias that makes it clear what the extension actually is since it's got a stupid name.
```php
use DNADesign\Elemental\TopPage\DataExtension as TopPageExtension;
```

GuySartorelli (changes requested):
Works well locally, just a couple of small changes.

src/Extensions/UsedOnTableExtension.php
emteknetnz: > Note that while the exists() call is currently redundant

Should just deprecate the method then rather than using it?

src/Extensions/UsedOnTableExtension.php
emteknetnz: Should just deprecate exists() then?

src/Extensions/UsedOnTableExtension.php
emteknetnz: Updated

src/Extensions/UsedOnTableExtension.php
emteknetnz: Updated

src/Extensions/UsedOnTableExtension.php
emteknetnz: Updated

GuySartorelli (approved):
LGTM

Issue #148, opened 2023-12-20
Handle cases where there is no owner


emteknetnz:
As part of [saving the relation between an Owner and Link on Link creation](https://github.com/silverstripe/silverstripe-linkfield/issues/145) you can get into a state where the Owner has yet to be saved, such as creating a new Owner on a GridField.

In these instances, it probably makes sense to not render the LinkField component normally, and instead have a message in the LinkField along the lines of "Save the Owner first"

We can simply check to see if the ownerID prop in the LinkField component is `0` or not as a condition for showing the message

## PR
- https://github.com/silverstripe/silverstripe-linkfield/pull/159

GuySartorelli:
@emteknetnz Did you mean to open this in the linkfield repo?
I'll migrate it there

GuySartorelli:
PR merged

PR #159, opened 2024-01-09
ENH Show save record first text on unsaved owners

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-linkfield/issues/148

Looks like this on unsaved owners

image


client/src/components/LinkField/LinkField.js
```diff
@@ -1,5 +1,5 @@
 /* eslint-disable */
-import React, { useState, useEffect, createContext } from 'react';
+import React, { useState, useEffect, createContext, Component } from 'react';
```
ChloeSartorelli: What is this change for?

client/src/components/LinkField/LinkField.js
```diff
@@ -211,6 +215,8 @@ const mapDispatchToProps = (dispatch) => ({
   },
 });
 
+export { LinkField as Component };
```
ChloeSartorelli, line 217: What is this change for? Can the test not use the default export?

client/lang/en.js
```diff
@@ -13,6 +13,7 @@ if (typeof(ss) === 'undefined' || typeof(ss.i18n) === 'undefined') {
     "LinkField.DELETE_ERROR": "Failed to delete link",
     "LinkField.ADD_LINK": "Add Link",
     "LinkField.ARCHIVE": "Archive",
-    "LinkField.DELETE": "Delete"
+    "LinkField.DELETE": "Delete",
+    "LinkField.SAVE_RECORD_FIRST": "Save record first"
```
ChloeSartorelli:
```suggestion
    "LinkField.SAVE_RECORD_FIRST": "Save the record before adding links"
```
This is a message for a content author rather than a developer, so it pays to be a little more verbose.
You'll need to update it everywhere else too.

ChloeSartorelli: I'm on the fence between this and "Cannot add links until the record has been saved", so take your pick between those (or some variation of those)

GuySartorelli (changes requested):
Works well locally - just a small change to the text and some questions about some small changes.

client/src/components/LinkField/LinkField.scss
```diff
@@ -0,0 +1,3 @@
+.link-field__save-record-first {
+  padding-top: 7px;
```
ChloeSartorelli, line 2: Convert padding to `rem` units, to match other padding in the CMS.

GuySartorelli (changes requested):

client/src/components/LinkField/LinkField.scss
emteknetnz: I won't change this. `px` is much more common, and far more intuitive

Doing a regex search in `./vendor/silverstripe/**/*.scss` for:

`padding.+rem` - 52 results in 14 files

`padding.+px` - 168 results in 45 files

ChloeSartorelli: Fair enough. I only looked at elements in the DOM near this element. Given there's a mix I guess it doesn't matter too much (until we go and change the font size but then this won't be the only area that needs fixing so it's a bit moot)

client/lang/en.js
emteknetnz: Updated

client/src/components/LinkField/LinkField.js
emteknetnz: Removed

client/src/components/LinkField/LinkField.js
emteknetnz: Nope, because the default export gets wired up to redux and will throw a bunch of terminal warnings when you try to use jest. This is a standard pattern used for doing jest tests.

GuySartorelli (approved):
LGTM, works well locally.

Issue #110, opened 2023-11-02
All AJAX calls should handle all possible errors gracefully


maxime-rainville:
## Story

As a content author, I want UI feedback when an HTTP request fail so I can take appropriate action with certainty.

## Notes
- Fairly sure that there are red toasts for failure already, we just need to double check this is the case

## Acceptance criteria
- All Linkfield AJAX requests are reviewed so they can handled error scenarios.
- Other unforeseen errors fail gracefully:
  - Server response timeout
  - Unexpected 500
  - Any other errors not already covered.

## PR
- https://github.com/silverstripe/silverstripe-linkfield/pull/161

GuySartorelli:
PR merged

PR #161, opened 2024-01-10
ENH Gracefully handle AJAX failures

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-linkfield/issues/110

I tested by adding the following in to controller endpoints on LinkFieldController:
- `$this->jsonError(500, 'FFF');` - expected failure
- `throw new Exception('AAA');` - unexpected failure
- `set_time_limit(1);while(true);` - simulate timeout

I tested when:
- Reading links `LinkFieldController::getLinkData()`
- Deleting links `LinkFieldController::linkDelete()`
- Saving links `LinkFieldController::save()`


client/lang/en.js
```diff
@@ -13,6 +13,13 @@ if (typeof(ss) === 'undefined' || typeof(ss.i18n) === 'undefined') {
     "LinkField.DELETE_ERROR": "Failed to delete link",
     "LinkField.ADD_LINK": "Add Link",
     "LinkField.ARCHIVE": "Archive",
-    "LinkField.DELETE": "Delete"
+    "LinkField.DELETE": "Delete",
+    "LinkField.LINK_DRAFT_TITLE": "Link has draft changes",
+    "LinkField.LINK_DRAFT_LABEL": "Draft",
+    "LinkField.LINK_MODIFIED_TITLE": "Link has unpublished changes",
```
emteknetnz, line 19: There was a bunch of translations missing from the "dist" version of en.js, so copy pasted them in from the "src" version

ChloeSartorelli: This isn't really necessary - when we finally do the standardisation of this module it'll be added to tx-translator's module set and we can do it all that way. This change is okay here but in future probably just leave it for when we're at that stage.

GuySartorelli:
There are conflicts to resolve.

emteknetnz:
Conflicts resolved

GuySartorelli (approved):
LGTM.

In a perfect world we'd add some JEST tests for this but I'm not too bothered about it, the main thing IMO is that the data doesn't end up in a bad state which it looks like it would be pretty hard to accomplish via thrown exceptions etc.

Issue #111, opened 2023-11-02
Provide a graceful loading animation for the LinkField


maxime-rainville:
## Story
As a content author, I want clear feedback from the UI so I can understand what elements are ready to use and what elements are still initialising.

## Note
- Ideally just use the existing loading component. People might use the js injector to replace the loading component, so it would be nice if they don't have to replace _multiple_ loading components.

## Acceptance criteria
- Linkfield shows the `<Loading>` component when appropriate
- Have tested using bandwidth restrictor using browser developer tools

## Related
- This [commit in AnyField](https://github.com/maxime-rainville/silverstripe-anyfield/commit/af8843bee3181729c778fe2f783db29965117065) how you can provide pre-react-render loading animation
- That's the AnyField animation. Maxime just took the regular loading animation and pop it into the LinkField Box. ![image](https://github.com/silverstripe/silverstripe-linkfield/assets/1168676/38646f90-3408-4a87-a7a4-4dfa70ee0991)

## PR
- https://github.com/silverstripe/silverstripe-linkfield/pull/162

sabina-talipova:
PR is merged. Close issue.

PR #162, opened 2024-01-10
ENH Show loading component during ajax read requests

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-linkfield/issues/111

I took inspiration from the "related" section in the parent issue for showing an 'inline' loading component so that it didn't put the overlay over the entire screen


sabina-talipova (approved):
Test locally. Works perfect.

Issue #160, opened 2024-01-09
Invalid prop type


emteknetnz:
When `yarn dev` in run in vendor/silverstripe/admin, the following error shows in console

<img width="483" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNFN4QWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--0dc806e429a4245dbae3ee4d0cab9d01b99f61c8/image.png" alt="image.png" />

## PR
- https://github.com/silverstripe/silverstripe-linkfield/pull/163

PR #163, opened 2024-01-10
FIX Console warning about required type

sabina-talipova:
## Description
- Updated property required type for props `types`.
- ~~If `$this->getOwner()->canEdit()` returns boolean value, then `$attributes['data-can-create']` is populated with `'data-can-create'` instead of boolean value. Now return string value.~~

## Parent issue
- https://github.com/silverstripe/silverstripe-linkfield/issues/160


src/Form/LinkField.php
```diff
@@ -44,7 +44,7 @@ protected function getDefaultAttributes(): array
     {
         $attributes = parent::getDefaultAttributes();
         $attributes['data-value'] = $this->Value();
-        $attributes['data-can-create'] = $this->getOwner()->canEdit();
+        $attributes['data-can-create'] = $this->getOwner()->canEdit() ? 'true' : 'false';
```
emteknetnz: I don't like the use of string booleans here, it gets confusing.  Revert this and just do things how Silverstripe does them natively

src/Form/MultiLinkField.php
```diff
@@ -61,7 +61,7 @@ protected function getDefaultAttributes(): array
     {
         $attributes = parent::getDefaultAttributes();
         $attributes['data-value'] = $this->getValueArray();
-        $attributes['data-can-create'] = $this->getOwner()->canEdit();
+        $attributes['data-can-create'] = $this->getOwner()->canEdit() ? 'true' : 'false';
```
emteknetnz: Revert so that we're not using string booleans

client/src/entwine/LinkField.js
```diff
@@ -53,8 +53,8 @@ jQuery.entwine('ss', ($) => {
         ownerRelation: inputField.data('owner-relation'),
         onChange: this.handleChange.bind(this),
         isMulti: this.data('is-multi') ?? false,
-        types: this.data('types') ?? [],
-        canCreate: this.getInputField().data('can-create') ?? false,
+        types: this.data('types') ?? {},
```
emteknetnz, line 56: Revert so that we're not using string booleans

emteknetnz (changes requested):
Note: I haven't tested locally

Should revert the using of string booleans because they're confusing and instead just rely on native Silverstripe behaviour.  Seems like the actual bug is coming from the use of `types` being an array instead of an object

src/Form/LinkField.php
sabina-talipova: Updated

src/Form/MultiLinkField.php
sabina-talipova: Updated

client/src/entwine/LinkField.js
sabina-talipova: I leave this part with new changes.

sabina-talipova: If `canEdit()` returns true, then `data-can-create` attribute exists and `inputField.data('can-create')`  = `data-can-create`, if `canEdit()` returns false, then `inputField.data('can-create')`  = `undefined`. So we can use this in JS to return `true` instead of return a string `data-can-create`.

emteknetnz (approved):

emteknetnz (changes requested):
Need to rebase to fix merge conflicts

sabina-talipova:
Rebased.

emteknetnz (approved):
Tested locally works good

PR #164, opened 2024-01-10
MNT Use FQCN for PHP translations

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-linkfield/issues/140

JS short classname namespace remains as it's the standard used in other modules e.g. admin

### DIFF
--- a/lang/en.yml
+++ b/lang/en.yml
@@ -1,44 +1,88 @@
 en:
-  LinkField:
-    ANCHOR_DESCRIPTION: 'Do not prepend "#". Anchor suggestions will be displayed once the linked page is attached.'
-    ANCHOR_FIELD_TITLE: 'Anchor'
+  SilverStripe\LinkField\Controllers\LinkFieldController:
     BAD_DATA: 'Bad data'
     CREATE_LINK: 'Create link'
-    DATA_HAS_NO_TYPEKEY: '"{class}": $data does not have a typeKey.'
     EMPTY_DATA: 'Empty data'
-    EXTERNAL_URL_FIELD: 'External url'
-    EMAIL_FIELD: 'Email address'
-    FILE_FIELD: 'File'
-    INVALID_DATA_TO_ARRAY: '"{class}": Could not convert $data to an array.'
-    INVALID_ID': 'Invalid ID'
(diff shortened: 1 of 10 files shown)


sabina-talipova:
Small question, should we also update en.yml file? Or will it have been updated automatically?

emteknetnz:
Good point. I've just run the text collector task on linkfield and committed the results

lang/en.yml
```diff
@@ -1,44 +1,88 @@
...
-    QUERY_FIELD_TITLE: 'Query string'
-    QUERY_STRING_DESCRIPTION: 'Do not prepend "?". EG: "option1=value&option2=value2"'
-    RECORDS_INSERTED: "{numrecords} records inserted, finished processing `{table}`\r\n"
-    TITLE_DESCRIPTION: 'Auto generated from Page title if left blank'
-    UNAUTHORIZED: 'Unauthorized'
+    MENUTITLE: SilverStripe\LinkField\Controllers\LinkFieldController
+    UNAUTHORIZED: Unauthorized
     UPDATE_LINK: 'Update link'
-    VERSIONED_STATUS_MISMATCH: 'Linkable and LinkField do not have matching Versioned applications. Make sure that both are either un-Versioned or Versioned'
+  SilverStripe\LinkField\Form\Traits\AllowedLinkClassesTrait:
```
ChloeSartorelli, line 16: Change references to these in the trait from `__CLASS__` to be the FCQN of the trait hardcoded, because:

https://www.php.net/manual/en/language.constants.magic.php
> When used in a trait method, __CLASS__ is the name of the class the trait is used in. 

In other words the call to `_t()` will pass in the name of the class (not the trait) - but here the key is the trait (not the class)

GuySartorelli (changes requested):

lang/en.yml
emteknetnz: Oh good spot :-)

Updated to use `__TRAIT__` (have confirmed it works as expected)

ChloeSartorelli: OH, nice! I should have looked at that list I linked to a little longer, I didn't notice that option.

GuySartorelli (approved):
LGTM

Issue #139, opened 2023-12-13
Add sort functionality to `MultiLinkField`


GuySartorelli:
The `MultiLinkField` can't be sorted right now - we should add that functionality.

## Acceptance Criteria
- Content authors can click-and-drag links in the `MultiLinkField` to determine their sort order
- The default sort order for `Link` is the sort field (meaning links are automatically in the correct order in the front-end)
- Sort order is saved on click/drag.
- JS sorting library is chosen with care towards using a well maintained library and minimise our maintenance burden.
- UX team is consultated on the design.

## Notes
- The AnyField implementation provided this functionality, so we might be able to pull some of this from there.

## PR
- https://github.com/silverstripe/silverstripe-linkfield/pull/166

maxime-rainville:
That's what the AnyField UI looks like right now. The handle comes from elemental.
![image](https://github.com/silverstripe/silverstripe-linkfield/assets/1168676/c25476ee-a30e-41ac-b83e-ef0614f1f3ed)

GuySartorelli:
PR merged

PR #166, opened 2024-01-10
ENH MutliLinkField sorting

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-linkfield/issues/139

### react-dnd -> dndkit

Most recent version of react-dnd is 16 which was released in Aug 2022 (we're still using version 5 on elemental), it failed to work at all in our setup
- I followed the basic "quick start" instructions and got a console warning
- Jest failed to work, complained about an 'Unexected export token' in a react-dnd file.  Apparently it uses the wrong JS syntax for our setup.
- Looking around some github comments it seems like the export issue started with version 16, it wasn't a problem with version 15

I've opted to use https://docs.dndkit.com/ instead which seems to work well and had its last release in Nov 2023

### Screen jumping on sort
The screen jumps on sort, though it also does this on anything that causes a re-render e.g. link edit.  I've created a new issue for this. https://github.com/silverstripe/silverstripe-linkfield/issues/169

### Hover/focus styling
Styling for the LinkField hover/active/focus states is generally poor, I've created a new issue to tidy this up https://github.com/silverstripe/silverstripe-linkfield/issues/168

### Design review
Design team has reviewed and approved with understanding the screen jumping and hover/focus styling will be handled separately


client/src/components/LinkPicker/LinkPicker.scss
```diff
@@ -66,11 +66,14 @@
   }
 
   &:hover, &:focus {
-    background: $gray-100;
```
emteknetnz, line 74: Styling for LinkField was a real mess, I've removed it because it was super obnoxious

I've created a new card to properly handle hover states https://github.com/silverstripe/silverstripe-linkfield/issues/168

GuySartorelli (changes requested):
The bottom border of the last item in the list is missing:
Before:
image

This PR:
image

GuySartorelli:
I took this after rebasing so I don't have the scroll popping issue.

[Screencast from 16-01-24 10:00:44.webm](https://github.com/silverstripe/silverstripe-linkfield/assets/36352093/63c35011-9d04-4134-bedd-13306ba9d5d3)

There are two issues here. 

## Flickering

I think we shouldn't bother trying to show the loading component here because that flickering is just very visually distracting and doesn't have any clear benefit.

At some point that decision looks like it was made for elemental as well, because elemental doesn't show the loading component when sorting.

## Briefly rendered in the wrong order

Very briefly after letting go of the item I'm dragging, it pops back into its _original_ position, and then pops into its _new_ position after loading.

We should immediately render it in its new position after dropping, and only pop it back to its original position if something goes wrong with the network request to save the new sort order.

client/src/components/LinkField/LinkField.js
```diff
@@ -190,7 +243,22 @@ const LinkField = ({
...
+      >
+        <SortableContext 
+          items={linkIDs}
+          strategy={verticalListSortingStrategy}
+        >
+          {links}
+        </SortableContext>
+      </DndContext>
+    </div> }
+    { !isMulti && <div>{ renderLinks() }</div>}
```
ChloeSartorelli: Put all of this into a new function (with a traditional `if` statement) to keep things nice and tidy - it's quite hard to read scan-read this code as it is here.
Probably call it something like `renderLinksList()`

```suggestion
    { !isMulti && <div>{ links }</div>}
```
I presume this was the point of turning it into a variable

client/src/components/LinkPicker/LinkPicker.scss
ChloeSartorelli: I was originally going to say not to remove this since hover states are going to be treated separately - but removing that is actually a genuine improvement.

src/Controllers/LinkFieldController.php
```diff
@@ -197,7 +202,10 @@ public function save(array $data, Form $form): HTTPResponse
         }
 
         // Ensure that ItemID url param matches the ID in the form data
-        if (isset($data['ID']) && ((int) $data['ID'] !== $id)) {
+        // Ensure that Sort is not being passed in
+        if ((isset($data['ID']) && ((int) $data['ID'] !== $id))
+            || isset($data['Sort'])
```
ChloeSartorelli, line 209: Why do we need this?
What if someone wants `Sort` to be a field that can be edited in the modal for their project, so they add it in?

src/Controllers/LinkFieldController.php
```diff
@@ -253,6 +261,62 @@ public function save(array $data, Form $form): HTTPResponse
...
+    /**
+     * Update Link Sort fields based frontend drag and drop sort logic
+     * /admin/linkfield/sort
+     */
+    public function linkSort()
+    {
+        $request = $this->getRequest();
+        // Check security token
+        if (!SecurityToken::inst()->checkRequest($request)) {
+            $this->jsonError(400, _t(__CLASS__ . '.INVALID_TOKEN', 'Invalid CSRF token'));
```
ChloeSartorelli, line 275: It feels weird telling probably-someone-doing-something-they-shouldn't what they got wrong... do we give such verbose responses about invalid tokens anywhere other than linkfield?

src/Controllers/LinkFieldController.php
```diff
@@ -253,6 +261,62 @@ public function save(array $data, Form $form): HTTPResponse
...
+        }
+        $json = json_decode($request->getBody() ?? '', true);
+        $newLinkIDs = (array) $json['newLinkIDs'];
+        if (empty($newLinkIDs)) {
+            $this->jsonError(400, _t('LinkField.BAD_DATA', 'Bad data'));
+        }
+        // Update index so that it can be used for the value of $sort
+        $newLinkIDs = array_values($newLinkIDs);
```
ChloeSartorelli: Can you please explain to me why this is needed? A JSON array will always be 0-indexed with clean keys anyway, won't it?

src/Controllers/LinkFieldController.php
```diff
@@ -253,6 +261,62 @@ public function save(array $data, Form $form): HTTPResponse
...
+        array_unshift($newLinkIDs, null);
+        unset($newLinkIDs[0]);
```
ChloeSartorelli: Why are we adding `null` and then removing it? This feels like a hacky way to do something - though I'm not sure what it's doing?

src/Controllers/LinkFieldController.php
```diff
@@ -253,6 +261,62 @@ public function save(array $data, Form $form): HTTPResponse
...
+        // Fetch and validate links
+        $links = Link::get()->filter(['ID' => $newLinkIDs])->toArray();
+        $linkIDToLink = [];
+        $ownerID = null;
+        $ownerRelation = null;
+        foreach ($links as $link) {
+            if (!$link->canEdit()) {
+                $this->jsonError(403, _t(__CLASS__ . '.UNAUTHORIZED', 'Unauthorized'));
+            }
```
ChloeSartorelli: If the link we can't edit isn't one of the links changing sort order, we shouldn't throw the error.
e.g:
- we have 4 links
- the 4th (last) link in the list can]t be edited
- we swap the top two links around

In that scenario we aren't editing the 4th link, so we don't care if we can edit it or not.

src/Controllers/LinkFieldController.php
```diff
@@ -253,6 +261,62 @@ public function save(array $data, Form $form): HTTPResponse
...
+            if (is_null($ownerID)) {
```
ChloeSartorelli, line 291: Some day we should really standardise `is_null` vs `=== null` (I think the latter is clearer) - but for now we haven't standardised that so I'll leave this alone :p

ChloeSartorelli, line 291:
```suggestion
            // Validate that all links are in the same relation with the same owner
            // This protects against malicious actors manually executing requests.
            if (is_null($ownerID)) {
```
Just add a comment saying why we're doing this - 'cause these won't happen under normal execution.

src/Controllers/LinkFieldController.php
```diff
@@ -253,6 +261,62 @@ public function save(array $data, Form $form): HTTPResponse
...
+                $ownerID = $link->OwnerID;
+            } elseif ($link->OwnerID !== $ownerID) {
+                $this->jsonError(400, _t('LinkField.BAD_DATA', 'Bad data'));
+            }
+            if (is_null($ownerRelation)) {
+                $ownerRelation = $link->OwnerRelation;
+            } elseif ($link->OwnerRelation !== $ownerRelation) {
+                $this->jsonError(400, _t('LinkField.BAD_DATA', 'Bad data'));
+            }
```
ChloeSartorelli:
```suggestion
            if (is_null($ownerID)) {
                $ownerID = $link->OwnerID;
                $ownerRelation = $link->OwnerRelation;
            } elseif ($link->OwnerID !== $ownerID || $link->OwnerRelation !== $ownerRelation) {
                $this->jsonError(400, _t('LinkField.BAD_DATA', 'Bad data'));
            }
```
Simplifying this since it's really part of the same check.

src/Controllers/LinkFieldController.php
```diff
@@ -253,6 +261,62 @@ public function save(array $data, Form $form): HTTPResponse
...
+        }
+        // Update Sort field on links
+        foreach ($newLinkIDs as $sort => $linkID) {
+            $linkID = $newLinkIDs[$sort];
+            $link = $linkIDToLink[$linkID];
+            if ($link->Sort !== $sort) {
+                $link->Sort = $sort;
+                $link->write();
+            }
+        }
```
ChloeSartorelli: If the stuff in https://github.com/silverstripe/silverstripe-linkfield/pull/166#discussion_r1452757193 and https://github.com/silverstripe/silverstripe-linkfield/pull/166#discussion_r1452757429 so that this `$sort` is 1-indexed:
1. Why 1-index it? What's wrong with 0-indexing?
2. If there's a really good reason to 1-index it:
    - just change `$sort` here to `$index` and have a `$sort = $index + 1` because that's much clearer than what's going on up there.
    - And please add a code comment here explaining why it's 1-indexed so when we see it in the future we don't need to hunt around to see what's going on.

src/Controllers/LinkFieldController.php
```diff
@@ -271,6 +335,9 @@ private function createLinkForm(Link $link, string $operation): Form
         $ownerClassName = $owner->ClassName;
         $ownerRelation = $this->ownerRelationFromRequest();
 
+        // Remove Sort field
+        $form->Fields()->removeByName('Sort');
```
ChloeSartorelli: Should remove in `getCMSFields()` of `Link` instead.
```suggestion
```

src/Models/Link.php
```diff
@@ -46,6 +47,8 @@ class Link extends DataObject
         ],
     ];
 
+    private static $default_sort = '"Sort"';
```
```suggestion
    private static $default_sort = 'Sort';
```
Shouldn't need `"` here iirc. It's not passed into a raw SQL method.

src/Models/Link.php
```diff
@@ -159,6 +162,14 @@ public function onBeforeWrite(): void
             $this->forceChange();
         }
 
+        // Esnure a Sort value is set
```
```suggestion
        // Ensure a Sort value is set
```

src/Models/Link.php
```diff
@@ -159,6 +162,14 @@ public function onBeforeWrite(): void
...
+        if (!$this->Sort) {
+            $this->Sort = self::get()->filter([
+                'OwnerID' => $this->OwnerID,
+                'OwnerRelation' => $this->OwnerRelation,
+            ])->max('Sort') + 1;
+        }
```
ChloeSartorelli, line 174: What's the intention here? Please add a comment to the code so it's clear going forward why we want this.

GuySartorelli (changes requested):

GuySartorelli:
Also needs rebase please

src/Controllers/LinkFieldController.php
emteknetnz: It's to prevent a malicious integer overflow attack, have added comment

src/Controllers/LinkFieldController.php
emteknetnz: We in in CSRFMiddleware in graphql, though beyond Graphql I don't think we do.  This can be useful to have this sort of thing as a dev message for developers.  I don't it makes things any easier for malicious users to do bad things so I think it's net positive.  All of LinkFieldController is already like this so best to keep as is for consistency. Was probably overkill making all the error messages translatable though :-)

src/Controllers/LinkFieldController.php
emteknetnz: Won't if you pass it in as an associate array defining integer indexes

src/Controllers/LinkFieldController.php
emteknetnz: It's so that we get a 1-index instead of a 0-index.  Basically it's more intuitive and standard to have a Sort value of 1 rather than 0.  Elemental also uses a 1-index

I've added a comment about this.

src/Controllers/LinkFieldController.php

src/Controllers/LinkFieldController.php
emteknetnz: Updated

src/Controllers/LinkFieldController.php

src/Controllers/LinkFieldController.php
emteknetnz: Already made a comment about about the 1-index
Have changed to `$sort = $i + 1`

src/Controllers/LinkFieldController.php

src/Models/Link.php
emteknetnz: Updated

src/Models/Link.php
emteknetnz: Updated

src/Models/Link.php
emteknetnz: It's so that newly created Links on MultiLinkFields are properly sorted. Have added comment

src/Controllers/LinkFieldController.php
ChloeSartorelli: I still don't understand this... can you please provide either some code so I can do this myself and see what happens (so I can see first hand what we're trying to avoid), or else explain further?

emteknetnz: If you set 'Sort' = the max value of integer in the POST request, you'll no longer be able to create links on that relation because of the ->max('Sort') + 1 logic in LinkField::onBeforeWrite()

client/src/components/LinkPicker/LinkPicker.js
```diff
@@ -4,7 +4,6 @@ import React, { useState } from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
 import LinkPickerMenu from './LinkPickerMenu';
-import LinkType from 'types/LinkType';
```
emteknetnz, line 7: This is unrelated refactoring to remove an unused import

package.json
```diff
@@ -1,16 +1,16 @@
 {
-  "name": "silverstripe-link",
+  "name": "silverstripe-linkfield",
```
emteknetnz, line 2: This is unrelated refactoring to use the correct repo name

src/Controllers/LinkFieldController.php
```diff
@@ -25,20 +25,24 @@ class LinkFieldController extends LeftAndMain
...
 
-    private static $allowed_actions = [
+    private static array $allowed_actions = [
         'linkForm',
         'linkData',
         'linkDelete',
+        'linkSort',
     ];
 
+    private static string $required_permission_codes = 'CMS_ACCESS_CMSMain';
```
emteknetnz, line 44: This is required to make the controller accessible to non-admin users, came up during design review

client/src/components/LinkField/LinkField.js
ChloeSartorelli: Wasn't done, but it either looks tidier now or my brain has gotten used to it from seeing it a couple times. Either way, I'll allow this to remain not done.

src/Controllers/LinkFieldController.php
ChloeSartorelli: Okay, that makes sense.

~~It doesn't look like that would do anything here though? Since `$data` isn't actually directly used anywhere, other than specifically for `$data['File']['Files'][0]` which is obviously not going to include anything in `$data['Sort']`.~~

Ahh nevermind, `$form` would also contain that data and `$form->saveInto($link);` would save it in.
This feels like a bit of an odd hack for a super unlikely edgecase to me - but that said, I don't think it's likely to bite us in the butt either having it nor omitting it, and since it's here now it may as well stay.

src/Controllers/LinkFieldController.php
ChloeSartorelli: > I don't it makes things any easier for malicious users to do bad things

I'd argue that saying "You have an invalid CSRF token" is making it easier for malicious users to do bad things - they forgot the token, or had a bad token, and we're telling them that's what was wrong with their request which will make it really obvious what they need to change to perform a successful attack.

I don't feel super strongly about this so I'll let you decide if you want to leave it in, or if that context is enough to make you want to remove it.

src/Controllers/LinkFieldController.php
ChloeSartorelli: But the array we're dealing with is an array from the JSON body content of a request. Which we're sending as an array.

I'm guessing that you're trying to protect against someone making a manual request which has a JSON *object*, which would resolve here as an associative array?

If that's the case, I think it's a lot more sensible to let JSON objects resolve to `stdclass`, and then check `if (!is_array($newLinkIDs)) $this->jsonError(400);` 'Cause as soon as someone's doing that, it's bad data that we shouldn't try to _guess_ what it should have been (which resetting the array with `array_values()` is doing).

src/Controllers/LinkFieldController.php
```diff
@@ -253,6 +263,69 @@ public function save(array $data, Form $form): HTTPResponse
...
+     */
+    public function linkSort()
+    {
+        $request = $this->getRequest();
+        // Check security token
+        if (!SecurityToken::inst()->checkRequest($request)) {
+            $this->jsonError(400, _t(__CLASS__ . '.INVALID_TOKEN', 'Invalid CSRF token'));
+        }
+        $json = json_decode($request->getBody() ?? '', true);
+        $newLinkIDs = (array) $json['newLinkIDs'];
```
ChloeSartorelli:
```suggestion
        $json = json_decode($request->getBody() ?? '', false);
        // If someone's passing a JSON object or other non-array here, they're doing something wrong
        if (!is_array($newLinkIDs)) {
            $this->jsonError(400);
        }
        
        $newLinkIDs = $json->newLinkIDs;
```
See discussion in https://github.com/silverstripe/silverstripe-linkfield/pull/166#discussion_r1452757193

client/src/components/LinkPicker/LinkPicker.scss
```diff
@@ -58,20 +58,28 @@
...
+  border-top-right-radius: 0;
+  border-bottom: 0;
+  border-bottom-left-radius: 0;
+  border-bottom-right-radius: 0;
+
+  &--is-first,
+  &--is-sorting {
+    border-top: 1px solid $gray-200;
+    border-top-left-radius: 0.23rem;
+    border-top-right-radius: 0.23rem;
```
ChloeSartorelli: Split these into their own `.link-picker__link--is-first,  .link-picker__link--is-sorting {` blocks.
When looking to find out where css is coming from, doing a code search for the full css class (e.g. search for `link-picker__link--is-first`).

Using the `&--` syntax makes this difficult, and unless the person doing the search thinks to check for that, they simply won't find the css in their search.

GuySartorelli (changes requested):
The actual sorting works really well now - I like that it's constrained to the parent and the vertical axis.

But now when I add new links, they don't appear until I refresh the page:
[Screencast from 17-01-24 15:03:47.webm](https://github.com/silverstripe/silverstripe-linkfield/assets/36352093/caa6285c-7cdc-4d06-92d9-ae430b2ef7f8)

src/Controllers/LinkFieldController.php
emteknetnz: Leave it in

client/src/components/LinkPicker/LinkPicker.scss
emteknetnz: Updated

src/Controllers/LinkFieldController.php

GuySartorelli (approved):
Looks great, works well locally.

Issue #40, opened 2021-11-15
Make link type labels and order consistent with WYSIWYG


purplespider:
When adding a link to a page this module uses the terminology "Site Tree Link" in the UI, but I don't believe the term "Site Tree" is used anywhere else in the CMS.

For clarity and consistency can we make the terminology the same as when inserting a link via the WYSIWYG, e.g.:
<img width="183" alt="Screenshot 2021-11-15 at 11 16 49@2x" src="https://user-images.githubusercontent.com/329880/141772950-2ddffbe0-c665-4291-9d06-48406a01a858.png">

Current linkfield module for comparison:
<img width="173" alt="Screenshot 2021-11-15 at 11 17 11@2x" src="https://user-images.githubusercontent.com/329880/141772994-464b5dc7-0a8b-446d-a1d0-defc944f455d.png">

(I started trying to do this myself, but unfortunately I don't do React).

## Acceptance criteria
- The labels in the LinkField drop down are updated to match the equivalent label in the WYSIWYG drop down
  - "SiteTree Link" is updated to be "Page on this site".
- The Link titles remain unchanged everywhere else.
- The options in the LinkField drop down are displayed in the same order that they're in with the WYSIWYG drop down

## PR
- https://github.com/silverstripe/silverstripe-linkfield/pull/167

purplespider:
@maxime-rainville How can I get some eyes on this? Do you agree? Any potential complications?

maxime-rainville:
We're doing a bunch of work on this right now. I've added this to the backlog.

GuySartorelli:
PR merged

PR #167, opened 2024-01-11
ENH Add sorting by link type

sabina-talipova:
## Description
- New class method that returns Menu labels;
- Add sorting for main Link classes;
 
## Parent issue
- https://github.com/silverstripe/silverstripe-linkfield/issues/40

### DIFF
--- a/src/Form/Traits/AllowedLinkClassesTrait.php
+++ b/src/Form/Traits/AllowedLinkClassesTrait.php
@@ -93,10 +93,14 @@ trait AllowedLinkClassesTrait
             }
             $typesList[$key] = [
                 'key' => $key,
-                'title' => $type->i18n_singular_name(),
+                'title' => $type->getMenuTitle(),
                 'handlerName' => $type->LinkTypeHandlerName(),
+                'priority' => $class::config()->get('menu_priority'),
             ];
         }
+        uasort($typesList, function ($a, $b) {
+            return $a['priority'] - $b['priority'];
+        });
 
         return json_encode($typesList);
     }
@@ -121,7 +125,6 @@ trait AllowedLinkClassesTrait
                 $result[$type] = $class;
             }
         }
-
         return $result;
     }
 }
--- a/src/Models/FileLink.php
+++ b/src/Models/FileLink.php
@@ -19,6 +19,11 @@ class FileLink extends Link
         'File' => File::class,
     ];
 
+    /**
+     * Set the priority of this link type in the CMS menu
+     */
+    private static int $menu_priority = 10;
+
     public function getCMSFields(): FieldList
     {
         $this->beforeUpdateCMSFields(function (FieldList $fields) {
@@ -54,5 +59,14 @@ class FileLink extends Link
         }
 
         return (string) $this->getDescription();
+    }
+
+    /**
+     * The title that will be displayed in the dropdown
+     * for selecting the link type to create.
+     */
+    public function getMenuTitle(): string
+    {
+        return _t(__CLASS__ . '.LINKLABEL', 'Link to a file');
     }
 }
--- a/src/Models/Link.php
+++ b/src/Models/Link.php
@@ -57,6 +57,11 @@ class Link extends DataObject
      */
     private ?string $linkType = null;
 
+    /**
+     * Set the priority of this link type in the CMS menu
+     */
+    private static int $menu_priority = 100;
+
     public function getDescription(): string
     {
         return '';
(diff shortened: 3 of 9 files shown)


GuySartorelli (changes requested):
This seems like unnecessary added complexity... why not just change the singular and plural names via `private static $singular_name` and `private static $plural_name`?

Edit: Nevermind, that would violate the acceptance criterion "The Link titles remain unchanged everywhere else." - so this is probably the right way to go about it.

src/Models/Link.php
```diff
@@ -454,4 +454,12 @@ public function getShortCode(): string
     {
         return strtolower(rtrim(ClassInfo::shortName($this), 'Link')) ?? '';
     }
+
+    /**
+     * Create menu title from singular_name
```
```suggestion
     * The title that will be displayed in the dropdown for selecting the link type to create.
     * Subclasses should override this - but it will use the singular_name by default.
```

src/Models/FileLink.php
```diff
@@ -55,4 +55,12 @@ public function getDefaultTitle(): string
 
         return (string) $this->getDescription();
     }
+
+    /**
+     * Create menu title for File link type
```
ChloeSartorelli:
```suggestion
     * The title that will be displayed in the dropdown for selecting the link type to create.
```
Use this same text for all except `Link` itself.

src/Form/Traits/AllowedLinkClassesTrait.php
```diff
@@ -121,7 +121,31 @@ private function genarateAllowedTypes(): array
...
+
+    private function sortLinkTypes(array $types): array
+    {
+        $order = [
+            'sitetree',
+            'file',
+            'external',
+            'email',
+            'phone',
+        ];
```
ChloeSartorelli: Add a `private static int $menu_priority = 100` config property to `Link` instead. Note the default value of 100 will be higher than any of our inbuilt links, which means custom links will by default fall to the bottom of the list.
The lower the priority value, the higher up in the list the link will be.

This makes it easy for custom links to be anywhere in the list, and for developers to reorder the list however they may choose.

Then set the following values (note going up in increments of 10 makes it easy for custom links to be place inbetween any links here)
|Link model|Priority value|
|---|---|
|`SiteTreeLink`|0|
|`FileLink`|10|
|`ExternalLink`|20|
|`EmailLink`|30|
|`PhoneLink`|40|

You can then order the array based on the priority config value.
The way I recommend doing that is to remove this sort method, and in `getTypesProps()` add an extra item to the typeslist arrays:

```diff
             $typesList[$key] = [
                 'key' => $key,
                 'title' => $type->getMenuTitle(),
                 'handlerName' => $type->LinkTypeHandlerName(),
+                'priority' => $class::config()->get('menu_priority'),
             ];
```

And then before encoding the array, use [uasort](https://www.php.net/manual/en/function.uasort.php) to sort it
```php
uasort($typesList, fn($a, $b) => $a['priority'] - $b['priority']);
return json_encode($typesList);
```


NOTE: There are other ways to do the sorting - the main thing I care about here is the configurable priority.

GuySartorelli (changes requested):

src/Models/FileLink.php
sabina-talipova: Updated

src/Models/Link.php

src/Form/Traits/AllowedLinkClassesTrait.php

src/Form/Traits/AllowedLinkClassesTrait.php
```diff
@@ -93,10 +93,14 @@ public function getTypesProps(): string
...
                 'key' => $key,
-                'title' => $type->i18n_singular_name(),
+                'title' => $type->getMenuTitle(),
                 'handlerName' => $type->LinkTypeHandlerName(),
+                'priority' => $class::config()->get('menu_priority'),
             ];
         }
+        uasort($typesList, function ($a, $b) {
+            return $a['priority'] - $b['priority'];
+        });
```
ChloeSartorelli, line 103:
```suggestion
        uasort($typesList, fn ($a, $b) => $a['priority'] - $b['priority']);
```
I won't block merging the PR for this - but in the future we should use the short function syntax for this sort of thing, since we're just returning a value.
The long function syntax is only really valuable when we've got multiple lines of functionality or when the return statement is really long.

GuySartorelli (approved):
linkfield with the link types in the new order with new labels
LGTM.

Issue #123, opened 2023-11-29
Allow link types to have custom icons


GuySartorelli:
If link types had their own icons, it would add some quick visual distinction between links of different types, which will help content authors quickly understand what they're looking at.

## Notes

Max had a PR where this was being implemented: https://github.com/silverstripe/silverstripe-linkfield/pull/45 - should be able to pick some stuff from this
![image](https://github.com/silverstripe/silverstripe-linkfield/assets/1168676/ff0aecee-e2e0-47b4-9633-c319b55569c3)

## ACs
- There are different icons for different links types
- Documentation shows how to add custom icons to custom link subclasses

## PR
- https://github.com/silverstripe/silverstripe-linkfield/pull/170

GuySartorelli:
@sabina-talipova please link to the PR(s) in the issue description

sabina-talipova:
> @sabina-talipova please link to the PR(s) in the issue description

Updated

GuySartorelli:
PR merged

PR #170, opened 2024-01-11
NEW Link type icon

sabina-talipova:
## Description
Screenshot 2024-01-12 at 12 09 39 PM

#### Note
Probably, we should have special icons for link types, especially for Phone link and File Link . Currently, icons from general Silverstripe library were used.  

## Parent issue
- https://github.com/silverstripe/silverstripe-linkfield/issues/123


src/Models/PhoneLink.php
```diff
@@ -19,6 +19,11 @@ class PhoneLink extends Link
      * Set the priority of this link type in the CMS menu
      */
     private static int $menu_priority = 40;
+    
+    /**
+     * Set the default icon for the Phone link type
+     */
+    private static $icon = 'font-icon-block-phone';
```
ChloeSartorelli:
```suggestion
    private static $icon = 'font-icon-mobile';
```
This is fairly subjective - but I think the mobile icon is easier to see (the box around the block-phone icon makes the actual phone icon really small).
the "block-phone" and "mobile" icons side-by-side

src/Models/EmailLink.php
```diff
@@ -23,6 +23,11 @@ class EmailLink extends Link
      */
     private static int $menu_priority = 30;
 
+    /**
+    * Set the default icon for the Email link type
+    */
```
ChloeSartorelli:
```suggestion
```
There's no need to specifically indicate that this is "for the Email link type" since that's implied by this being the `EmailLink` class. Just let it inherit the PHPDoc from `Link`.

Same with all other subclasses of `Link`.

src/Models/Link.php
```diff
@@ -62,6 +62,11 @@ class Link extends DataObject
      */
     private static int $menu_priority = 100;
 
+    /**
+     * Set the default icon for the link type
+     */
```
```suggestion
    /**
     * The css class for the icon to display for this link type
     */
```
It's not directly an icon - it's a css class.
And it's not the "default" css class, it's _the_ css class for the icon.

src/Models/Link.php
```diff
@@ -470,4 +475,9 @@ public function getMenuTitle(): string
     {
         return $this->i18n_singular_name();
     }
+
+    public function getLinkIcon(): string
+    {
+        return static::config()->get('icon');
+    }
```
```suggestion
```
There's no need to have this method - just use `::config()->get('icon')` directly.

src/Form/Traits/AllowedLinkClassesTrait.php
```diff
@@ -96,6 +96,7 @@ public function getTypesProps(): string
                 'title' => $type->getMenuTitle(),
                 'handlerName' => $type->LinkTypeHandlerName(),
                 'priority' => $class::config()->get('menu_priority'),
+                'icon' => $type->getLinkIcon(),
```
ChloeSartorelli:
```suggestion
                'icon' => $class::config()->get('icon'),
```

client/src/components/LinkField/LinkField.js
```diff
@@ -164,6 +164,7 @@ const LinkField = ({
         description={data[linkID]?.description}
         versionState={data[linkID]?.versionState}
         typeTitle={type.title || ''}
+        typeIcon={type.icon || 'font-icon-link'}
```
ChloeSartorelli: I don't think this `||` condition needs to be in here, since the icon will fall back to `font-icon-link` as defined on the `Link` class anyway.
```suggestion
        typeIcon={type.icon}
```

README.md
```diff
@@ -138,6 +138,30 @@ use SilverStripe\Versioned\Versioned;
 Link::remove_extension(Versioned::class);
 ```
 
+## Additional features
+
+The developer can customize the position of the link type in the menu by setting the `$menu_priority` value. The priority is set ascending, i.e. the link will be displayed in the menu above if the priority value is less. By default, the highest priority has SiteTreeLink equal to zero.
```
ChloeSartorelli:
```suggestion
The developer can customise the position of the link type in the menu by setting the `$menu_priority` value. The priority is in ascending order (i.e. a link with a higher priority value will be displayed lower in the list).
```
"customise" instead of "customize" - the z is american english.
I've reworded this to use much simpler language, and I've removed the bit about `SiteTreeLink` equaling zero because people can just look at the config in the code to see that.

README.md
```diff
@@ -138,6 +138,30 @@ use SilverStripe\Versioned\Versioned;
...
+The developer can also set an icon that will correspond to a specific type of link by setting the value of the `$icons` property. This name must correspond to the name of the class of the icon that needs to be used.
```
ChloeSartorelli:
```suggestion
The developer can also set an icon that will correspond to a specific type of link by setting the value of the `$icon` configuration property. The value of this configuration corresponds to the css class of the icon to be used.
```

GuySartorelli (changes requested):
Works well locally, just a couple of small changes.

src/Models/PhoneLink.php
sabina-talipova: Updated

src/Models/EmailLink.php
sabina-talipova: Updated

src/Models/Link.php

src/Models/Link.php

src/Form/Traits/AllowedLinkClassesTrait.php

client/src/components/LinkField/LinkField.js
sabina-talipova: Updated

README.md
sabina-talipova: Updated

README.md
sabina-talipova: Updated

GuySartorelli (approved):
LGTM

Issue #169, opened 2024-01-11
Screen jumps up on rerender of MultiLinkField


emteknetnz:
When you have a number of Links in a MultiLinkField and the LinkField react component rerenders creating an AJAX request to refetch Links, and for a short time the amount of vertical space taken up by the component reduces and it causes the browser screen to abruptly move up, loosing the users place on the screen

https://www.youtube.com/watch?v=HPimduM7bRU

This is because we render the loading component _instead of_ the list of links.
We should instead wash out the link list and display the loading component _on top of_ it. This will retain all of the rendered content while we're loading (preventing this scroll jump), and makes it a lot clearer _what_ specifically is loading.

## PRs
- https://github.com/silverstripe/silverstripe-linkfield/pull/171

PR #171, opened 2024-01-15
FIX Don't change scroll position when loading

GuySartorelli:
The issue was caused by not rendering all of the elements that make up the list of links in the field while loading - which meant they're removed from the DOM, which adjusts the total height of the window.
Once the items are added back to the DOM, the window height adjusts again and therefore the user is not at the same scroll point they were at before the load event occurred.

This is fixed by following the pattern `dnadesign/silverstripe-elemental` uses - we render the loading component _on top_ of the existing list of items.

Screenshot of some link fields and an elemental area, all loading

The buttons in the link fields can still be accessed by tabbing to them, so I've disabled them when we're loading.

## Issue
- https://github.com/silverstripe/silverstripe-linkfield/issues/169


client/src/components/LinkField/tests/LinkField-test.js
```diff
@@ -71,7 +71,10 @@ test('LinkField will render link-picker if ownerID is not 0 and has finished loa
   })}
   />);
   doResolve();
-  await screen.findByText('Add Link');
+  // Short wait - we can't use screen.find* because we're waiting for something to be removed, not added to the DOM
+  await act(async () => {
+    await new Promise((resolve) => setTimeout(resolve, 100));
+  });
```
ChloeSartorelli, line 76: The `act()` here and in other tests is needed to avoid a warning:

> Warning: An update to LinkField inside a test was not wrapped in act(...).
>       
> When testing, code that causes React state updates should be wrapped into act(...):
>       
>       act(() => {
>         /* fire events that update state */
>       });
>       /* assert on the output */
>       
> This ensures that you're testing the behavior the user would see in the browser. Learn more at https://reactjs.org/link/wrap-tests-with-act

client/src/components/LinkPicker/tests/LinkPicker-test.js
```diff
@@ -1,39 +1,59 @@
 /* global jest, test */
 import React from 'react';
-import { render } from '@testing-library/react';
+import { render, fireEvent } from '@testing-library/react';
+import { act } from 'react-dom/test-utils';
```
emteknetnz:
```suggestion
import { render, fireEvent, act } from '@testing-library/react';
```

client/src/components/LinkField/tests/LinkField-test.js
```diff
@@ -1,6 +1,7 @@
 /* global jest, test */
 import React from 'react';
-import { render, screen } from '@testing-library/react';
+import { render } from '@testing-library/react';
+import { act } from 'react-dom/test-utils';
```
emteknetnz:
```suggestion
import { render, act } from '@testing-library/react';
```

emteknetnz (changes requested):
Test locally, works good, could of minor import changes

client/src/components/LinkPicker/tests/LinkPicker-test.js
ChloeSartorelli: Ahh, good catch. Didn't know it was available through that library as well.

ChloeSartorelli: Done

client/src/components/LinkField/tests/LinkField-test.js
ChloeSartorelli: Done

emteknetnz (approved):

PR #172, opened 2024-01-16
ENH Add Readonly field status

sabina-talipova:
## Description
The developer can set readonly using the `setReadonly` method and in the CMS the user will not be able to edit or delete record data. The user will only be able to view these records.
Also, this state will be used if the user can only view the page, but not edit it.

Screenshot 2024-01-16 at 1 43 05 PM

### Test steps 
 - Add the following code to your test DataObject or Page:
 
 ```php
        'HasOneLink' => Link::class,
    ];

    private static $has_many = [
        'HasManyLinksOne' => Link::class . '.Owner',
        'HasManyLinksTwo' => Link::class . '.Owner',
    ];

    private static array $owns = [
        'HasOneLink',
        'HasManyLinksOne',
        'HasManyLinksTwo',
    ];

    {
        $fields = parent::getCMSFields();
        $fields->removeByName(['HasOneLinkID', 'HasManyLinksOne', 'HasManyLinksTwo']);

        $fields->addFieldsToTab(
            'Root.Main',
            [
                LinkField::create('HasOneLink')
                ->setReadonly(true),
                MultiLinkField::create('HasManyLinksOne'),
                MultiLinkField::create('HasManyLinksTwo')
                ->setReadonly(true),
            ],
        );

        return $fields;
    }
 ```

#### Note
There is some problem here, how this should work if the user can edit the page, but `LinkField` is set to read only. Which should take precedence over `canEdit` or `ReadOnly`? Since if the field is read only, then a user with `canEdit` permission will not be able to create a link. Should there be a super user who can create links? Based on this discussion: https://github.com/silverstripe/silverstripe-linkfield/pull/141#discussion_r1433406199

## Parent issue
- https://github.com/silverstripe/silverstripe-linkfield/issues/11


src/Controllers/LinkFieldController.php
```diff
@@ -316,6 +321,19 @@ private function createLinkForm(Link $link, string $operation): Form
...
 
+    /**
+     * Get is the owner LinkField is readonly
+     */
+    private function isReadOnlyField(): bool
+    {
+        $request = $this->getRequest();
+        $ownerClass = $request->getVar('ownerClass') ?: $request->postVar('OwnerClass');
+        $ownerRelation = $this->ownerRelationFromRequest();
+        $isReadOnly = Injector::inst()->get($ownerClass)->getCMSFields()->dataFieldByName($ownerRelation)?->isReadonly();
```
sabina-talipova: This method gets Owner's (e.g. Page) fields and check is this LinkField is read only.

src/Controllers/LinkFieldController.php
```diff
@@ -316,6 +321,19 @@ private function createLinkForm(Link $link, string $operation): Form
...
+
+        return $isReadOnly ?? false;
```
sabina-talipova: `isReadOnly ` can be `null` in some cases.

GuySartorelli:
I'm not doing an actual review of this yet - just answering the question in the note:
> There is some problem here, how this should work if the user can edit the page, but `LinkField` is set to read only. Which should take precedence over `canEdit` or `ReadOnly`? Since if the field is read only, then a user with `canEdit` permission will not be able to create a link. Should there be a super user who can create links? Based on this discussion: https://github.com/silverstripe/silverstripe-linkfield/pull/141#discussion_r1433406199

Readonly is more important than `canEdit`. It doesn't matter if we can edit anything if the field has been set to readonly - readonly explicitly means we're not allowed to edit the field, even if we have edit permissions for the model or the model's owner.

client/src/components/LinkPicker/LinkPicker.scss
```diff
@@ -105,6 +105,10 @@
   &--published::before {
     display: none;
   }
+
+  &.readonly {
```
emteknetnz:
```suggestion
  &--readonly {
```
Safer to use a BEM class rather than a generic/global `readonly` that may accidentally get values set from somewhere else

You'll need to update this is bunch of places in this PR

client/src/components/LinkPicker/LinkPickerTitle.js
```diff
@@ -56,17 +58,17 @@ const LinkPickerTitle = ({
...
+          {getVersionedBadge(versionState)}
+        </div>
+        <small className="link-picker__type">
+          {typeTitle}:&nbsp;
+          <span className="link-picker__url">{description}</span>
+        </small>
       </div>
     </Button>
-    {canDelete &&
+    {(canDelete && canCreate) &&
```
emteknetnz: This logic looks wrong, shouldn't have a `canCreate` check for a delete button

src/Controllers/LinkFieldController.php
```diff
@@ -449,6 +467,11 @@ private function ownerRelationFromRequest(): string
         if (!$ownerRelation) {
             $this->jsonError(404, _t(__CLASS__ . '.INVALID_OWNER_RELATION', 'Invalid ownerRelation'));
         }
+
+        if (strpos($ownerRelation, '_Readonly') !== false) {
```
emteknetnz: Why are we doing this with $ownerRelation?  Owner relation should be what's set in the private static $has_many, not the `_Readonly` added to form fields when they are set to read only mode?

src/Controllers/LinkFieldController.php
```diff
@@ -316,6 +321,19 @@ private function createLinkForm(Link $link, string $operation): Form
         return $form;
     }
 
+    /**
+     * Get is the owner LinkField is readonly
+     */
+    private function isReadOnlyField(): bool
+    {
+        $request = $this->getRequest();
+        $ownerClass = $request->getVar('ownerClass') ?: $request->postVar('OwnerClass');
```
emteknetnz: Create a new private method getOwnerClassFromRequest() so that it's consistent with the rest of the file and throw a 404 error if the owner class !is_a DataObject

src/Controllers/LinkFieldController.php
```suggestion
        return (bool) Injector::inst()->get($ownerClass)->getCMSFields()->dataFieldByName($ownerRelation)?->isReadonly();
```

src/Controllers/LinkFieldController.php
```suggestion
```

emteknetnz (changes requested):
I haven't tested locally yet

I really dislike the how 'readonly' has been stuffed into the value of 'canCreate'.  It makes things much harder to understand what's going on.

Instead you should make 'readonly' its own separate key / prop

client/src/components/LinkPicker/LinkPickerTitle.js
sabina-talipova: Updated. I have to use `readonly`, since we should hide "Archive" button if user cannot edit link field.

client/src/components/LinkPicker/LinkPicker.scss
sabina-talipova: Updated

src/Controllers/LinkFieldController.php
sabina-talipova: Updated

src/Controllers/LinkFieldController.php
sabina-talipova: Updated

src/Controllers/LinkFieldController.php
sabina-talipova: Updated

src/Controllers/LinkFieldController.php
sabina-talipova: Updated

src/Controllers/LinkFieldController.php
ChloeSartorelli: This doesn't take into account scenarios where the _form_ was set to readonly, which I believe is what happens when the user doesn't have edit permission for the owner record (e.g. a page).

src/Controllers/LinkFieldController.php
```diff
@@ -376,9 +384,9 @@ private function createLinkForm(Link $link, string $operation): Form
         });
 
         // Make readonly if fail can check
-        if ($operation === 'create' && !$link->canCreate()
-            || $operation === 'edit' && !$link->canEdit()) {
+        if ($isNotEditable) {
             $form->makeReadonly();
+            $form->setMessage(_t(__CLASS__ . '.READ_ONLY_MESSAGE', 'You do not have enough permissions to edit this form'), 'info');
```
emteknetnz: I'm not sure we should be setting a form message here. Can you find any other examples of us doing this?

If we do keep it, the message is wrong when the form has been manually set to read only rather than because of a lack of permissions

client/src/components/LinkPicker/tests/LinkPickerTitle-test.js
```diff
@@ -89,3 +91,17 @@ test('LinkPickerTitle main button should not fire the onClick callback while loa
   fireEvent.click(container.querySelector('button.link-picker__button'));
   expect(mockOnClick).toHaveBeenCalledTimes(0);
 });
+
+test('LinkPickerTitle render() should have readonly class if cannot edit', () => {
```
emteknetnz:
```suggestion
test('LinkPickerTitle render() should have readonly class if set to readonly', () => {
```

client/src/components/LinkPicker/tests/LinkPickerTitle-test.js
```diff
@@ -89,3 +91,17 @@ test('LinkPickerTitle main button should not fire the onClick callback while loa
...
+  const { container } = render(<LinkFieldContext.Provider value={{ loading: false }}>
+    <LinkPickerTitle {...makeProps({ readonly: true })} />
+  </LinkFieldContext.Provider>);
+  expect(container.querySelectorAll('.link-picker__link--readonly')).toHaveLength(1);
+});
+
+test('LinkPickerTitle render() should not have readonly class if can edit', () => {
```
emteknetnz:
```suggestion
test('LinkPickerTitle render() should not have readonly class if set to readonly', () => {
```

emteknetnz (changes requested):
I'm still able to sort links in a MultiLinkField when it's in a readonly state

Also needs a rebase

client/src/components/LinkPicker/tests/LinkPickerTitle-test.js
sabina-talipova: Updated

client/src/components/LinkPicker/tests/LinkPickerTitle-test.js
sabina-talipova: Updated

src/Controllers/LinkFieldController.php
sabina-talipova: @emteknetnz, I could not find any similar example. I think it was new requirements from Design team. I moved logic only for readonly field.

src/Controllers/LinkFieldController.php
```diff
@@ -377,16 +378,32 @@ private function createLinkForm(Link $link, string $operation): Form
 
         // Make readonly if fail can check
         if ($operation === 'create' && !$link->canCreate()
-            || $operation === 'edit' && !$link->canEdit()) {
+            || $operation === 'edit' && !$link->canEdit()
+            || $this->isReadOnlyField()) {
             $form->makeReadonly();
         }
 
+        if ($this->isReadOnlyField()) {
```
emteknetnz: This message feels even more out of place now

It will now only show if `$this->isReadOnlyField()` is explicitly called, though not if `$this->canEdit()` is false or if the owner page/element `canEdit()` returns false which seems worse because it's inconsistent.

We also don't show this message anywhere else in the CMS as far as we know

I think we should just remove this message, it's seems pretty obvious that the form is readonly because the background of all the form elements if grey

Perhaps raise a new card to add this to globally add this message to forms through the CMS instead?

emteknetnz (changes requested):

src/Controllers/LinkFieldController.php
sabina-talipova: Updated

emteknetnz (changes requested):
Merge conflicts, needs a rebase + run yarn build

emteknetnz (approved):

Issue #165, opened 2024-01-10
Handle trailing slashes in urls


emteknetnz:
We have javascript like this that assumes no trailing slashes in the url

`${Config.getSection(section).form.linkForm.deleteUrl}/${linkID}`;

Need to use the url helper that Guy made to join these safely

GuySartorelli:
PR merged

PR #173, opened 2024-01-16
ENH Join endpoint path segments

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-linkfield/issues/165


client/src/components/LinkField/LinkField.js
```diff
@@ -2,16 +2,15 @@
 import React, { useState, useEffect, createContext } from 'react';
 import { bindActionCreators, compose } from 'redux';
 import { connect } from 'react-redux';
-import { injectGraphql } from 'lib/Injector';
```
emteknetnz, line 5: This is minor unrelated refactoring, I removed a couple of unused imports

client/src/components/LinkField/LinkField.js
```diff
@@ -114,7 +113,7 @@ const LinkField = ({
    * Update the component when the 'Delete' button in the LinkPicker is clicked
    */
   const onDelete = (linkID) => {
-    let endpoint = `${Config.getSection(section).form.linkForm.deleteUrl}/${linkID}`;
+    let endpoint = joinUrlPaths(Config.getSection(section).form.linkForm.deleteUrl, linkID.toString());
```
emteknetnz, line 116: Need to convert the linkID int to a string otherwise joinURLPaths gets a weird array created when you pass in a string then an int 0 as arguments

GuySartorelli (approved):
Looks good, works well locally with both trailing slash config values. Can't find any that were missed.

Issue #151, opened 2023-12-21
Handle empty list of types


emteknetnz:
If there are no allowable types of links to be created the LinkPicker will end up with a fatal error e.g. 

`[Emergency] Uncaught InvalidArgumentException: "SilverStripe\LinkField\Form\LinkField": Allowed types cannot be empty`

This can be forced with `LinkField::create('MyLink')->setAllowedTypes([]);`, though it's also possible if the CMS user fails canCreate() checks on the allowed link types.

Since the CMS user has no control over this, we should allow and empty list of types and display a UI message on the frontend instead of an empty dropdown

## ACs
- There should be a message should be "Unable to create link"
- There should not be a fatal PHP error arriving via ajax
- If develop passes an empty array to setAllowedTypes() there should still be an exception because this is a developer error rather than a bunch of permission checks failing

## PR
- https://github.com/silverstripe/silverstripe-linkfield/pull/174

GuySartorelli:
@emteknetnz Please add a link to the PR

emteknetnz:
Done

sabina-talipova:
PR was merged. Close.

PR #174, opened 2024-01-17
ENH Show a cannot create message when types is empty

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-linkfield/issues/151

Test with
```php
    {
        // ...
        LinkField('MyLink')::create()->setAllowedTypes([MyLink::class]));
        // ...
```

Where MyLink has

```php
    public function canCreate($member = null, $context = [])
    {
        return false;
    }
```


src/Form/Traits/AllowedLinkClassesTrait.php
```diff
@@ -13,7 +13,9 @@
  */
 trait AllowedLinkClassesTrait
 {
-    private array $allowed_types = [];
+    private array $allowedTypes = [];
```
emteknetnz: Changing the casing in unrelated refactoring

client/src/components/LinkPicker/LinkPicker.js
```diff
@@ -40,12 +40,13 @@ const LinkPicker = ({ types, onModalSuccess, onModalClosed, canCreate }) => {
   const shouldOpenModal = typeKey !== '';
   const className = classnames('link-picker', 'form-control');
   const typeArray = Object.values(types);
+  const message = i18n._t('LinkField.CANNOT_CREATE_LINK', 'Cannot create link');
 
-  if (!canCreate) {
+  if (!canCreate || typeArray.length === 0) {
```
sabina-talipova, line 45: Probably, we should divide cases when user cannot create link and when there is not available link types.

sabina-talipova (changes requested):
I've tested locally, works perfect. Looks good, just small question.

client/src/components/LinkPicker/LinkPicker.js
emteknetnz: I'm not sure why we want to have separate logic here? The end result of both of seems like it should be the same, a message saying that you cannot create a link

What else would we do here if we split the logic?

src/Form/Traits/AllowedLinkClassesTrait.php
```diff
@@ -43,16 +46,6 @@ public function getAllowedTypes(): array
      */
     private function validateTypes(array $types): bool
     {
-        if (empty($types)) {
-            throw new InvalidArgumentException(
```
ChloeSartorelli: Still need to throw this exception if developers manually pass in an empty array - see acceptance criteria

emteknetnz: Ah yup my bad, didn't need to change any PHP in the end, have reverted

client/src/components/LinkPicker/LinkPicker.js
sabina-talipova: No, everything looks good to me.

sabina-talipova (approved):
LGTM.

PR #180, opened 2024-01-19
FIX Defaul link type title and icon for disabled link types

sabina-talipova:
## Parent issue
- https://github.com/silverstripe/silverstripe-linkfield/issues/176


GuySartorelli (changes requested):
If I set the `canCreate()` method for any link type to return `false`, it breaks again. Here's an example where `canCreate()` returns `false` for `PhoneLink`:
image

This PR also introduces a regression for https://github.com/silverstripe/silverstripe-linkfield/issues/151 
image

sabina-talipova:
> If I set the canCreate() method for any link type to return false, it breaks again. Here's an example where canCreate() returns false for PhoneLink,
> This PR also introduces a regression for https://github.com/silverstripe/silverstripe-linkfield/issues/151

Fixed

GuySartorelli (approved):
Looks good, works as expected.

PR #184, opened 2024-01-23
ENH Standardise API responses

emteknetnz:
Issue https://github.com/silverstripe/developer-docs/issues/422

- Use jsonSuccess() from https://github.com/silverstripe/silverstripe-admin/pull/1658
- Remove messages from error responses
- Use ~201~ 204 rather than 200 for empty body success responses

Requires https://github.com/silverstripe/silverstripe-admin/pull/1658 to be merged first for CI to pass. Once that is merged mark this PR as ready for review and rerun tests.

### DIFF
--- a/src/Controllers/LinkFieldController.php
+++ b/src/Controllers/LinkFieldController.php
@@ -74,17 +74,17 @@ class LinkFieldController extends LeftAndMain
         if ($id) {
             $link = Link::get()->byID($id);
             if (!$link) {
-                $this->jsonError(404, _t(__CLASS__ . '.INVALID_ID', 'Invalid ID'));
+                $this->jsonError(404);
             }
             $operation = 'edit';
             if (!$link->canView()) {
-                $this->jsonError(403, _t(__CLASS__ . '.UNAUTHORIZED', 'Unauthorized'));
+                $this->jsonError(403);
             }
         } else {
             $typeKey = $this->typeKeyFromRequest();
             $link = LinkTypeService::create()->byKey($typeKey);
             if (!$link) {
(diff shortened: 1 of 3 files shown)


src/Controllers/LinkFieldController.php
```diff
@@ -151,10 +147,7 @@ public function linkDelete(): HTTPResponse
             $owner->write();
         }
         // Send response
-        $response = $this->getResponse();
-        $response->addHeader('Content-type', 'application/json');
-        $response->setBody(json_encode(['success' => true]));
-        return $response;
+        return $this->jsonSuccess(201);
```
ChloeSartorelli: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/201
> The HTTP 201 Created success status response code indicates that the request has succeeded and has led to the creation of a resource.

This is for the deletion of the resource, not creation.
Either use [200](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/200)
> The HTTP 200 OK success status response code indicates that the request has succeeded.

or [204](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/204)
> The HTTP 204 No Content success status response code indicates that a request has succeeded, but that the client doesn't need to navigate away from its current page. 

Same with the sort method's response.

GuySartorelli (changes requested):

src/Controllers/LinkFieldController.php
emteknetnz: Ah my bad, confused 201 with 204

GuySartorelli (approved):
LGTM, works as expected locally.

Issue #181, opened 2024-01-19
Disabled state


emteknetnz:
Need to support a disabled state which is set via `->setDisabled(true)`

This is similar though not quite the same as `->setReadonly()` which had supported on [this card](https://github.com/silverstripe/silverstripe-linkfield/issues/11)

For example, with very quick test locally shows that there is a mouse pointer with the readonly field, though there isn't one with the disabled field

```php
    public function getCMSFields()
    {
        $fields = parent::getCMSFields();
        $fields->insertAfter('Content', new TextField('MyText01'));
        $fields->insertAfter('Content', (new TextField('MyText02'))->setReadonly(true));
        $fields->insertAfter('Content', (new TextField('MyText03'))->setDisabled(true));
        return $fields;
    }   
```

Some differences between readonly and disabled can be found [here](https://stackoverflow.com/questions/7730695/whats-the-difference-between-disabled-disabled-and-readonly-readonly-for-ht)

## Acceptance criteria
- When `setDisabled` is set to `true`, the field renders with the read only styles and the user can not update the value.
- There's a [follow up card](https://github.com/silverstripe/silverstripe-linkfield/issues/182) to block any updates in the backend for readonly/disabled field.

## Notes
- We discussed whether disabled or readonly should have any practical difference. We concluded it wasn't really worth treating them as separate use cases.

## PR
- https://github.com/silverstripe/silverstripe-linkfield/pull/189

sabina-talipova:
PR is merged. Task is completed.

PR #189, opened 2024-01-24
ENH Add disabled field state

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-linkfield/issues/181


sabina-talipova (changes requested):
@emteknetnz, it looks like we should modify [LinkFieldController::isReadOnlyField](https://github.com/silverstripe/silverstripe-linkfield/blob/f5d54ac9a91cfcbbd42daa421d3e41f055266745/src/Controllers/LinkFieldController.php#L395C22-L401) to check `LinkField->isDisable()`. Currently if we click on any disable link field we still can open modal window and edit link content.

emteknetnz:
Updated

src/Controllers/LinkFieldController.php
```diff
@@ -142,8 +143,8 @@ public function linkDelete(): HTTPResponse
         // delete() will also delete any published version immediately
         $link->delete();
         // Update owner object if this Link is on a has_one relation on the owner
-        $owner = $this->ownerFromRequest();
-        $ownerRelation = $this->ownerRelationFromRequest();
+        $owner = $this->getOwnerFromRequest();
+        $ownerRelation = $this->getOwnerRelationFromRequest();
```
emteknetnz, line 147: I've done some unrelated refactoring to ensure that all of the "ownerFrom*" methods are prefixed with "get" i.e. "getOwnerForm*"

Previously there was a mixture of methods with the prefix and those without

src/Controllers/LinkFieldController.php
```diff
@@ -20,7 +20,8 @@
 use SilverStripe\LinkField\Services\LinkTypeService;
 use SilverStripe\ORM\DataList;
 use SilverStripe\ORM\DataObject;
-use SilverStripe\Security\Security;
```
emteknetnz, line 23: This is unrelated refactoring to remove an unused import

src/Controllers/LinkFieldController.php
```diff
@@ -379,7 +380,7 @@ private function createLinkForm(Link $link, string $operation): Form
         // Make readonly if fail can check
         if ($operation === 'create' && !$link->canCreate()
             || $operation === 'edit' && !$link->canEdit()
-            || $this->isReadOnlyField()) {
+            || $this->getFieldIsReadonlyOrDisabled()) {
             $form->makeReadonly();
```
sabina-talipova, line 384: Now, it works as expected with one small issue, it use `readonly` attribute instead of `disabled` for all inputs. 
If you test here `$form->transform(new DisabledTransformation());` all input fields in modal window will get proper attribute `disabled`. Probably, the logic for disabled form should be extracted from readonly logic. 

And another small issue that element with class `.link-picker__link` has attribute `aria-disabled="false"`, but I assume it should be `aria-disabled="true"` or it should be removed.

sabina-talipova, line 384: Now, it works as expected with one small issue, it use `readonly` attribute instead of `disabled` for all inputs. 
If you test here `$form->transform(new DisabledTransformation());` all input fields in modal window will get proper attribute `disabled`(but unfortunately it doesn't work with DropdownField). Probably, the logic for disabled form should be extracted from readonly logic. 

And another small issue that element with class `.link-picker__link` has attribute `aria-disabled="false"`, but I assume it should be `aria-disabled="true"` or it should be removed.

sabina-talipova (changes requested):

src/Controllers/LinkFieldController.php
emteknetnz: In the issue notes it says:

"We discussed whether disabled or readonly should have any practical difference. We concluded it wasn't really worth treating them as separate use cases."

Basically disabled is such a narrow use case it's not worth investing much time getting this working exactly as it should.

Given that I don't think we should spend any time fixing either of these small issues

sabina-talipova (approved):
Test locally. All works perfect.

Issue #156, opened 2024-01-09
Cannot select existing file in FileLink with smaller than lg breakpoint


emteknetnz:
When selecting an existing file in FileLink, the LinkField shows 'File does not exist'.

This does not happen when uploading a new file.

![image](https://github.com/silverstripe/silverstripe-linkfield/assets/4809037/e6c7ce8e-4d1b-4408-9fb3-4c9c0562682b)

**Update: Use a breakpoint less than 992px wide, which is the lg breakpoint in _bottostrap-variables.scss**

## Follow up issue created
- https://github.com/silverstripe/silverstripe-admin/issues/1662

## PR
- https://github.com/silverstripe/silverstripe-linkfield/pull/190

sabina-talipova:
@emteknetnz , could you, please, provide steps to reproduce this issue. Test locally and didn't see any problem.

 
<img width="465" alt="Screenshot 2024-01-11 at 12 54 34 PM" src="https://github.com/silverstripe/silverstripe-linkfield/assets/87288324/122f997c-a692-4a7a-9978-643b16bbc68b">

emteknetnz:
1. Create a FileLink on either a has_one or has_many relationship
2. Upload a new file
3. Delete the FileLink
4. Refresh the page (otherwise you'll run into an asset-admin caching issue and you won't see the newly uploaded file)
5. Create a new FileLink
6. Select the recently uploaded file
7. LinkField says "File does not exist"

sabina-talipova:
Still see path as a File Link.
<img width="469" alt="Screenshot 2024-01-11 at 1 38 15 PM" src="https://github.com/silverstripe/silverstripe-linkfield/assets/87288324/483e281d-e897-48dc-ad33-e1d4ff904f86">

GuySartorelli:
I saw this happen once (in a fresh environment) - but when I tried to reproduce it (in that same environment), it didn't happen a second time. Seems like this one's gonna be a bit finicky.

emteknetnz:
This likely has something to do with the file selection modal getting into a weird state. You can replicate this by doing something unusual and clicking the 'details' button on a selected file and clicking "Save" on the file, rather than the "Insert" button.  When you do you don't go back to the LinkFieldModal with this 'Create/Update Link' button, instead you go straight back to the page edit form - likely it's this existing issue - https://github.com/silverstripe/silverstripe-linkfield/issues/114

[...]

PR #190, opened 2024-01-25
FIX Prevent nested asset-admin forms from submitting modal

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-linkfield/issues/156


client/src/components/LinkModal/LinkModal.js
```diff
@@ -35,6 +35,17 @@ const LinkModal = ({ typeTitle, typeKey, linkID = 0, isOpen, onSuccess, onClosed
...
   const onSubmit = async (modalData, action, submitFn) => {
     let formSchema = null;
+
+    // Workaround to prevent the FormBuilder form from submitting when any nested FormBuilder forms are present
+    // This will happen when a FileLink is being used and the "Choose existing" link is clicked which
+    // will open up an asset-admin modal to select an existing file.
+    // Clicking any of the following submit buttons will cause LinkModal to inadvertanely submit:
+    // - "Insert file" which is present when the breakpoint is less than 'lg' (992px)
+    // - Save/Publish button after clicking on the 'Details' tab of a selected file
+    if (document.querySelector('.asset-admin')) {
```
ChloeSartorelli: I was about to approve but I realised that this might prevent us (or anyone else) from adding linkfield to the edit fields of a file/folder inside asset-admin itself. We probably need something more specific than just "is that element anywhere at all in the DOM".

GuySartorelli (changes requested):
This is working around a wider issue, which is going to be handled properly in https://github.com/silverstripe/silverstripe-admin/issues/1662

For now this is an okay solution for linkfield

client/src/components/LinkModal/LinkModal.js
ChloeSartorelli: There might be information in the args being passed in here that we can use to determine which submit button was used (one of ours or one from asset-admin) - if so, that would be a better workaround.

emteknetnz: I couldn't find anything useful in the args, mainly because they're the same valid args as if you submitted the form modal normally.  It's just when they're inadvertently submitted by an asset-admin button the UploadField hasn't yet been populated.

We could check if the submitted uploadfield has a length of 0 files, however that's still a valid submission.  If we were to require the UploadField to have at least one file this is still the wrong place to be doing validation.

>  adding linkfield to the edit fields of a file/folder inside asset-admin itself.

I don't think that's a realistic concern.  There isn't a practical reason why anyone would add LinkField either of those, especially not a FileLink.

I've updated the file selector to be more precise as requested though

ChloeSartorelli: Turns out that while this level of specificity is better, linkfield doesn't work there because of unrelated react shenanigans anyway lol.
I still prefer this higher specificity regardless though. Thank you for making that change.

GuySartorelli (approved):
Works as expected

Issue #168, opened 2024-01-11
Tidy up hover/active/focus state styling


emteknetnz:
Hover/active/focus state styling is a bit of a mess

![image](https://github.com/silverstripe/silverstripe-linkfield/assets/4809037/c006a72b-8653-45bc-b389-a0bfbc703d32)

## Acceptance criteria
- Check with design team what they'd like done (probably just make everything have a white background and unchanging border to match elemental)
- Implement design suggestions

## PR
- https://github.com/silverstripe/silverstripe-linkfield/pull/191

sabina-talipova:
PR's merged. Task is completed.

PR #191, opened 2024-01-25
ENH Improve hover active and focus state styling

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-linkfield/issues/168

Has been demoed to and approved by design team


client/src/components/LinkPicker/LinkPickerTitle.js
```diff
@@ -95,10 +95,10 @@ const LinkPickerTitle = ({
           <span className="link-picker__url">{description}</span>
         </small>
       </div>
+      {(canDelete && !readonly) &&
+        <span className="link-picker__delete btn btn-link" onClick={stopPropagation(() => !loading ? onDelete(id) : null)}>{deleteText}</span>
```
emteknetnz: This has changed from a `<Button>` to a `<span>` because there's a error in yarn build that you cannot have a `<Button>` be a child of a `<Button>`

I've added the `btn btn-link` classes so that it still styles the same as the `<Button>`

sabina-talipova: Everything looks good, just small issue.
When we change `<button>` to `<span>` we lose all available button functionality. So now we lose focus on Archive button and when we use keyboard we cannot press "Archive" button.
Probably we should add additional attributes ([see](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role#basic_button_example)) and define onKeyDown method.
E.g. 
```js
<span role="button" tabIndex="0" className="link-picker__delete btn btn-link" onKeyDown={stopPropagation(() => !loading ? onDelete(id) : null)} onClick={stopPropagation(() => !loading ? onDelete(id) : null)}>{deleteText}</span>
```

sabina-talipova (changes requested):

client/src/components/LinkPicker/LinkPickerTitle.js

sabina-talipova (changes requested):
@emteknetnz, there are still some issues with "span like button":
- if we use keyboard and press "Archive" then select "Cancel" in confirmation window the Link edit modal window will be open after browser confirmation window.
- Small style issue with External links. 
Screenshot 2024-01-29 at 2 02 10 PM

client/src/components/LinkPicker/LinkPickerTitle.js
```diff
@@ -97,10 +101,23 @@ const LinkPickerTitle = ({
...
       </div>
+      {(canDelete && !readonly) &&
+        // This is a <span> rather than a <Button> because we're inside a <Button> and
+        // trigger an error when you attempt to nest a <Button> inside a <Button>.
+        <span
+          role="button"
+          tabIndex="0"
+          className="link-picker__delete btn btn-link"
+          onKeyDown={(evt) => {
+            preventDefaultPropogation(evt);
```
sabina-talipova: @emteknetnz , could you, please, update this method. Since we have a small issue with moving further.
Probably we should replace it with `stopPropagation` and add `evt.preventDefault()` after `onDelete(id)`

Before (Cannot select Button below)

https://github.com/silverstripe/silverstripe-linkfield/assets/87288324/4f4fa81a-a02b-4259-9e5d-065f1a49d0f5

After:

https://github.com/silverstripe/silverstripe-linkfield/assets/87288324/9b28bc61-dfb5-4f1c-b504-22fb90c908c4

sabina-talipova (changes requested):

sabina-talipova (approved):
LGTM. Tested locally, it works as expected.

Issue #152, opened 2023-12-19
Versioned UI state (of link owner record) after link creation and deletion


emteknetnz:
After a link is created and the underlying parent object is directly saved in the database, [as part of this card](https://github.com/silverstripe/silverstripe-linkfield/issues/145), the versioned UI state should be correct

Examples:
- For a freshly published Page with a LinkField on it, when a new Link is added, the Page versioned UI state should be a clean Save button, and a dirty Published button.
- For a freshly published Content Block Page with a LinkField on a Content Block, when a new Link is added, the Content Block should have a modified versioned state, and the page its only should have a clean Save button, and a dirty Published button.

## Notes
- Should be able to use the existing javascript change tracker to update this, rather than directly updating the save and publish buttons via jquery

## Acceptance criteria
- [ ] The versioned state UI for pages is correct when new Links are added or deleted
- [ ] The versioned state UI for content blocks is correct when new Links are added or deleted
- [ ] If there are any _unsaved_ changes on the page/block at the time the link is added/deleted, the save button should remain in a dirty state.

## PR
- https://github.com/silverstripe/silverstripe-linkfield/pull/193

## New issues created
- https://github.com/silverstripe/silverstripe-admin/issues/1665

sabina-talipova:
PR's merged. Task was completed.

PR #193, opened 2024-01-30
ENH Set Published button dirty state when Link is an unpublished state

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-linkfield/issues/152

I've tested this for:
- Links on Pages
- Links on Elemental
- Links on DataObjects managed by ModelAdmins

It will set the "Publish" button to a dirty state when the LinkPickerTitle has either a 'Draft' or a 'Modified' versionedState.  It won't change the state of "Save" button as LinkField constantly saves everything via AJAX.

Note that for Elemental this is only true if the Link if modifed via the link modal.  The "Save" button will still get a dirty state when new links are added or deleted.  This is due to existing logic in Elemental to that will trigger the jQuery changetracker which will alter both the Save and the Publish buttons.  There's nothing we can really do about this without a lot of hacking around it, and I don't think this is worth it since a dirty Save button is far from the worse thing ever.

I've also created a new issue for the all the CMS [inconsistency](https://github.com/silverstripe/silverstripe-admin/issues/1665) I discovered about the changetracker while doing this work


client/src/components/LinkField/LinkField.js
```diff
@@ -136,7 +137,7 @@ const LinkField = ({
   /**
    * Update the component when the 'Delete' button in the LinkPicker is clicked
    */
-  const onDelete = (linkID, deleteType) => {
+  const handleDelete = (linkID, deleteType) => {
```
emteknetnz, line 140: This is unrelated refactoring to follow the consistent naming of defining 'handle*' methods and passing them to 'on*' props

client/src/components/LinkField/LinkField.js
```diff
@@ -172,6 +173,42 @@ const LinkField = ({
...
+      '.cms-edit-form button[data-text-alternate]#Form_ItemEditForm_action_doPublish'
+    ].join(',');
+    const publishButton = document.querySelector(cssSelector);
+    if (!publishButton) {
+      return;
+    }
+    const dataBtnAlternateRemove = publishButton.getAttribute('data-btn-alternate-remove') || '';
+    dataBtnAlternateRemove.split(' ').forEach((className) => {
+      publishButton.classList.remove(className);
+    });
```
sabina-talipova: Everything works perfect, as expected. But there is a small issue here. If somehow we pass empty string as class name then we get `Uncaught DOMException: Failed to execute 'remove' on 'DOMTokenList': The token provided must not be empty.`. Probably we should check that `dataBtnAlternateRemove`, `dataBtnAlternateAdd `, `dataTextAlternate ` are not empty string before.

sabina-talipova (changes requested):

client/src/components/LinkField/LinkField.js
emteknetnz: Have added a couple of `if (className) {` checks

sabina-talipova (approved):
LGTM. Tested locally.

PR #195, opened 2024-01-30
Fix RequiredFields validation

emteknetnz:
Two PR fixes two things in seperarte PRs, so do not squash merge
1. Fixing non-inline editing of elements which was totally busted `private static bool $inline_editable = false;` on the content block
2. Adds the new HasOneRelationFieldInterface to LinkField, which means that it can be used by the RequiredFields validator 

Requires https://github.com/silverstripe/silverstripe-framework/pull/11119 to be merged first

### DIFF
--- a/src/Form/Traits/LinkFieldGetOwnerTrait.php
+++ b/src/Form/Traits/LinkFieldGetOwnerTrait.php
@@ -27,8 +27,12 @@ trait LinkFieldGetOwnerTrait
         $relation = $this->getName();
         // Elemental content block
         if (class_exists(BaseElement::class) && is_a($owner, BaseElement::class)) {
+            // Remove namespaces from inline editable blocks
+            // This will return an empty array for non-inline editable blocks (e.g. blocks in a gridfield)
             $arr = ElementalAreaController::removeNamespacesFromFields([$relation => ''], $owner->ID);
-            $relation = array_keys($arr)[0];
+            if (array_key_exists(0, $arr)) {
+                $relation = array_keys($arr)[0];
+            }
         }
         return [
             'ID' => $owner->ID,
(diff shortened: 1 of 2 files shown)


src/Form/Traits/LinkFieldGetOwnerTrait.php
```diff
@@ -27,8 +27,12 @@ private function getOwnerFields(): array
         $relation = $this->getName();
         // Elemental content block
         if (class_exists(BaseElement::class) && is_a($owner, BaseElement::class)) {
+            // Remove namespaces from inline editable blocks
+            // This will return an empty array for non-inline editable blocks (e.g. blocks in a gridfield)
             $arr = ElementalAreaController::removeNamespacesFromFields([$relation => ''], $owner->ID);
-            $relation = array_keys($arr)[0];
+            if (array_key_exists(0, $arr)) {
+                $relation = array_keys($arr)[0];
+            }
```
ChloeSartorelli:
```suggestion
            if (!empty($arr)) {
                $relation = array_keys($arr)[array_key_first($array)];
            }
```
This feels more robust - without this it seems like there's something magic going on where there might be other values in the array, but we only want values if it's 0-indexed, or something? 

But if we just want the first element (if there are any elements), this code should get us there without ambiguity.

That said, if we do explicitly want the 0-indexed value, we need a comment saying what the heck is goin on.

GuySartorelli (changes requested):
Just a small change.

src/Form/Traits/LinkFieldGetOwnerTrait.php
emteknetnz: Updated, though using slightly different `$relation = array_key_first($arr);` which is equivalent to how it was before

ChloeSartorelli: Yup, nice. I wasn't looking too closely, my code was garbage lol. Glad you got the idea.

GuySartorelli (approved):
LGTM, works well locally.

Issue #99, opened 2023-09-28
LinkField is accessible


maxime-rainville:
## Story
As visually impaired content author, I want to be able to manage Links.

## Some things to watch out for
- should be able to tab through links and press enter to open the modal, even in read-only
- opening the modal should give focus to the first field in the modal automatically
- if closing or submitting the modal using the keyboard, focus should be returned to the button which would open the modal for that link (or to the "add link" button for new link modal)
- Should be able to submit modal using the enter key when focus is on any text field
- Should be able to reorder links with the keyboard.
- Aria label should be present whenever an image/icon is used instead of text.

## Acceptance criteria
- Accessibility issues with LinkField are identified.
- Simple issues are addressed.
- Complex issues are spun up as separate card.

## Exception
- In some cases the LinkField maybe reusing other CMS UI elements that are not accessible (e.g. File Upload modal). Fixing those other elements is beyond the scope of this card.

## Note
- Since this is a one of the first time we do a proper accessibility review, we should take learning from this card into a retro topic.

## New issues created
- https://github.com/silverstripe/silverstripe-linkfield/issues/205

## PRs
- https://github.com/silverstripe/silverstripe-linkfield/pull/200

maxime-rainville:
Cassie said this should be a high priority.

emteknetnz:
> opening the modal should give focus to the first field in the modal automatically

This one is too hard for now.  Since the modal is not part of linkfield it should be enhanced in admin instead of doing something hackish in linkfield to get it to focus on the first.

In order to do this properly we'd need to create a ref via `useRef()` in FormBuilder for each field that's created, and pass that into the create form field component via an innerRef prop, and in EVERY one of react components add an innerRef prop and then put that ref on the relevant `<input>` or whatever.   We'd then have a new prop in form builder focusOnFirstField=false which can be set to true, which we'd do if opening the form builder modal via keyboard, and then formbuilder with go someRef.current.focus() on open.  This would be a signficant piece of work.

While this is nice to have, in practice all it does is save 2x tab presses to focus the first input

emteknetnz:
> Accessibility issues with LinkField are identified.

Going from https://govtnz.github.io/web-a11y-guidance/

**Accessible UX best practices:**
- Colour and contract - Out of scope for this card - broader issue with CMS that should be dealt with in a CMS redesign
- Designing documents for print - Not applicable for the CMS
[...]

PR #200, opened 2024-02-01
ENH Improve keyboard support

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-linkfield/issues/99

Does the following:
- Tab through links and press enter (or space) to open the modal, even in read-only
- If closing or submitting the modal using the keyboard, focus should be returned to the button which would open the modal for that link (or to the "add link" button for new link modal)
- Reorder links with the keyboard - tab to the link handle, press enter/space then press up/down or left/right, then press enter/space to accept
- Focuses on the first input when opening the link modal with keyboard

Jest testing proved extremely hard so I haven't done much of it in this PR. I switched to using React Testing Library [userEvent](https://testing-library.com/docs/user-event/intro/) though it still didn't behave like a browser, for instance:
- keyboard sorting isn't possible because jsdom doesn't have a layout engine so cannot calculated new coords
- pressing enter/space on a react strap dropdown item didn't trigger the `onclick` handler which is triggered when doing the same in an actual browser

New issue for behat keyboard sorting:
- https://github.com/silverstripe/silverstripe-linkfield/issues/205


client/src/components/LinkField/LinkField.js
```diff
@@ -42,7 +49,6 @@ const section = 'SilverStripe\\LinkField\\Controllers\\LinkFieldController';
 const LinkField = ({
   value = null,
   onChange,
-  onNonPublishedVersionedState,
```
emteknetnz, line 45: This is unrelated refactoring to remove a non-existant prop

client/src/components/LinkField/LinkField.js
```diff
@@ -1,11 +1,18 @@
...
+  DndContext,
+  KeyboardCode,
+  KeyboardSensor,
+  PointerSensor,
+  useSensor,
+  useSensors
+} from '@dnd-kit/core';
 import { arrayMove, SortableContext, verticalListSortingStrategy } from '@dnd-kit/sortable';
 import { restrictToVerticalAxis, restrictToParentElement } from '@dnd-kit/modifiers';
-import { injectGraphql } from 'lib/Injector';
```
emteknetnz, line 8: This is unrelated refactoring to remove an unused import

client/src/components/LinkPicker/LinkPicker.scss
```diff
@@ -139,19 +139,26 @@
...
   left: 5px;
   position: absolute;
   z-index: 100;
 
   &:hover {
     cursor: grab;
   }
+
+  .font-icon-drag-handle {
+    opacity: 0%;
```
emteknetnz, line 152: Using opacity instead of display so that it can be tabbed to

client/src/components/LinkPicker/LinkPickerTitle.js
```diff
@@ -108,15 +143,7 @@ const LinkPickerTitle = ({
           role="button"
           tabIndex="0"
           className="link-picker__delete btn btn-link"
-          onKeyDown={(evt) => {
```
emteknetnz, line 111: This is unrelated refactoring, moved to handleDeleteKeyDown()

client/src/components/LinkPicker/LinkPickerTitle.js
```diff
@@ -59,6 +61,27 @@ const LinkPickerTitle = ({
...
+  const handleButtonKeyDown = (event) => {
+    // Prevent the triggering the parent's keyboard sorting handler
+    event.nativeEvent.stopImmediatePropagation();
+    event.stopPropagation();
+    if (['Enter', 'Space'].includes(event.code) && !loading) {
+      onButtonKeyDownEdit(event);
+    }
+  };
+
+  const handleDeleteKeyDown = (event) => {
```
emteknetnz, line 74: This is unrelated refactoring, was previously an inline method in render() for onKeyDown

client/src/components/LinkPicker/LinkPicker.scss
sabina-talipova, line 1: I'm not sure if this is a big problem, but shouldn't we use the same style when focusing on linkfield elements? What is the opinion of the designers?
Screenshot 2024-02-09 at 8 47 15 AM
Screenshot 2024-02-09 at 8 57 36 AM

client/src/components/LinkPicker/LinkPickerTitle.js
```diff
@@ -82,15 +114,36 @@ const LinkPickerTitle = ({
...
+    { (isMulti && !readonly && !disabled) && <div className="link-picker__drag-handle">
+      <i
+        className="font-icon-drag-handle"
+        tabIndex="0"
+        role="button"
+        aria-pressed="false"
+        aria-controls={idAttr}
+        onKeyDown={handleIconKeyDown}
+      ></i>
+    </div> }
```
sabina-talipova:
```suggestion
    { (isMulti && !readonly && !disabled) && <div className="link-picker__drag-handle"
        tabIndex="0"
        role="button"
        aria-pressed="false"
        aria-controls={idAttr}
        aria-label="Sort Links"
        onKeyDown={handleIconKeyDown}
    >
      <i
        className="font-icon-drag-handle"
        aria-hidden="true"
        focusable="false"
      ></i>
    </div> }
```

sabina-talipova: Probably, would be better to use as a button `div` element instead of `icon`. 
And also, please, add `aria-label="Sort Links"` since this element doesn't have any content that explains what this button does.

client/src/components/LinkPicker/LinkPickerTitle.js
```diff
@@ -82,15 +114,36 @@ const LinkPickerTitle = ({
...
+    <Button
+      disabled={loading}
+      className={`link-picker__button ${typeIcon}`}
+      color="secondary"
+      onClick={stopPropagation(onClick)}
+      innerRef={buttonRef}
+      onKeyDown={handleButtonKeyDown}
```
sabina-talipova, line 151: This element also should have `aria-label` attribute with something like "Add link" or "Edit link".
Otherwise, screen reader reads it as something like this `DRAFT Phone number: 1234567833333 Archive`.
This is a good article about `area` attributes. https://www.sarasoueidan.com/blog/accessible-icon-buttons/

client/src/components/LinkPicker/LinkPickerTitle.js
```diff
@@ -108,15 +161,7 @@ const LinkPickerTitle = ({
...
-          onKeyDown={(evt) => {
-            if ((evt.code === 'Enter' || evt.code === 'Space') && !loading) {
-              evt.nativeEvent.stopImmediatePropagation();
-              evt.stopPropagation();
-              onDelete(id);
-              evt.nativeEvent.preventDefault();
-              evt.preventDefault();
-            }
-          }}
+          onKeyDown={handleDeleteKeyDown}
```
sabina-talipova, line 171:
```suggestion
          onKeyDown={handleDeleteKeyDown}
          aria-label={deleteText}
```

client/src/components/LinkField/LinkField.js
```diff
@@ -106,34 +169,96 @@ const LinkField = ({
...
+  // Ensure the exact same number of hooks are called on every render
+  // If this this isn't done then a react error will be thrown when a link is deleted
+  while (refCount < 256) {
+    useRef(null);
+    refCount++;
+  }
+
+  // Focus on the LinkPickerTitle edit button when the editing modal is closed via keyboard
+  // or focus on newly created link for single (has_one) linkfield
+  useEffect(() => {
```
sabina-talipova, line 188: Just small question. Double execution `useEffect` won't it be a reason for some problems with `setInterval`?

sabina-talipova (changes requested):
Looks good, works perfect. Just tiny possible improvements.

client/src/components/LinkField/LinkField.js
emteknetnz: No, it'll just create a separate reference to a new internal.

client/src/components/LinkPicker/LinkPicker.scss
emteknetnz: I talked to design about this, they said they're happy with this as it's a default browser style and there's no need to override this as it's obvious what it's doing

client/src/components/LinkPicker/LinkPickerTitle.js
emteknetnz: Updated

client/src/components/LinkPicker/LinkPickerTitle.js

sabina-talipova (changes requested):
@emteknetnz, could you please fix this one as well https://github.com/silverstripe/silverstripe-linkfield/pull/200/files#r1483616400

client/src/components/LinkPicker/LinkPickerTitle.js
emteknetnz: I added this on line 167

emteknetnz:
@sabina-talipova Have updated

sabina-talipova (approved):
LGTM. Test locally, works perfect.

Issue #183, opened 2024-01-22
Add LinkField to pattern library


maxime-rainville:
## Acceptance criteria
- Add Pattern library stories for Single and MultiLinkField 
- Linkfield has been added to the [Pattern library](https://silverstripe.github.io/silverstripe-pattern-lib)

## Note
- It's probably not possible to include LinkModals because they are driven by form schema which will not work in StoryBook.
- Versioned link states can be handled with dynamic props.

## PRs
- https://github.com/silverstripe/silverstripe-linkfield/pull/204
- https://github.com/silverstripe/silverstripe-pattern-lib/pull/4

PR #204, opened 2024-02-07
MNT Add to storybook

GuySartorelli:
I wasted most of a day trying to mock fetch requests, and failed miserably. Because of that, we can't show what the linkfield looks like when it's populated with links since that data is fetched asynchronously rather than being passed in as a prop.
We also can't render the link modal here for similar reasons.

## Issue
- https://github.com/silverstripe/silverstripe-linkfield/issues/183


emteknetnz (approved):
Tested locally, works great

Issue #203, opened 2024-02-07
"Archive/Delete" button should not be an option when field is disabled


GuySartorelli:
I can click the "Archive" (or "Delete" for unversioned) link when the field is disabled. That should not be possible.

## Acceptance criteria
- The "Archive"/"Delete" button is disabled (but still visible) when the field is disabled.

## PR
- https://github.com/silverstripe/silverstripe-linkfield/pull/206

GuySartorelli:
PR merged

PR #206, opened 2024-02-07
FIX Do not show delete button when disabled

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-linkfield/issues/203

On the initial [disabled card](https://github.com/silverstripe/silverstripe-linkfield/issues/181) we decided:

`We discussed whether disabled or readonly should have any practical difference. We concluded it wasn't really worth treating them as separate use cases.`

The readonly state will simply not show the archive/disable button rather than having it in a disabled state. Because of what was decided on the original disabled card I've opted to treat the disabled state exactly the same.


GuySartorelli (approved):
Makes sense. I don't necessarily agree with the decision that was originally made, but I'm okay with abiding with it given it has been made. This is better than the previous behaviour in any case.

Issue #10, opened 2020-11-23
Link versioned history


maxime-rainville:
## Story
As a **CMS User** I want **to view old link instance when looking at previous version of my page** so that **I can restore old content with confidence**.

## Acceptance criteria
- LinkField display in ReadOnly view when looking at an older version of the page instead of the "Can not create Link" view.
- Version of link displayed follows whatever logic versioned history follows for other relations.
- Archived Admin uses the same behaviour as Versionned Admin when displaying link.
- LinkField on versioned history view correctly identify if a Link was attached to the page, but the Link was deleted.
- LinkField on versioned history view correctly identify if a Link was not attached to the version of the page currently being viewed.

## Note
- We do not have an overarching view of how relations should be displayed in compare view. We will ignore this part for now.
- We considered adding a history tab to the LinkForm, but decided it was out of scope for this card.

## Designs
[Link field - Overview](https://projects.invisionapp.com/dsm/silver-stripe/silverstripe-cms/nav/build/folder/components/5fbc28d0096742ec25194c31?mode=preview)

## Note
The current implementation reuses the same IDs when you change Link Type. This could be a plus ... or a horrible hindrance ... not sure which one yet.

## PR
- https://github.com/silverstripe/silverstripe-linkfield/pull/209

sachajudd:
@silverstripeux to look into what this would look like in history.

maxime-rainville:
Cassie didn't think there was much value beyond making sure the LinkField is usable in the Versioned history view.

maxime-rainville:
## What the diff looks like

![image](https://github.com/silverstripe/silverstripe-linkfield/assets/1168676/4039e872-4ee9-4207-803b-ea10eecf8734)

## What the older version looks like
![image](https://github.com/silverstripe/silverstripe-linkfield/assets/1168676/f8b147c0-15f7-435e-9a10-8d407d26bb45)

GuySartorelli:
PR merged

PR #209, opened 2024-02-08
ENH Better versioned history

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-linkfield/issues/10

May require https://github.com/silverstripe/silverstripe-versioned-admin/issues/198 to be merged in order for archive admin to work properly

Note that for elemental page history where you have block with a has_one linkfield on it, the value for this nested field will not display.  This is consistent with nested file's not showing on the history of an elemental-file-block with a linked file on it which also does not show on the page history.  In both instances, clicking on the "Block history" link will show the value of nested field


client/src/components/LinkField/LinkField.js
```diff
@@ -5,7 +5,6 @@ import { connect } from 'react-redux';
 import { DndContext, closestCenter, PointerSensor, useSensor, useSensors } from '@dnd-kit/core';
 import { arrayMove, SortableContext, verticalListSortingStrategy } from '@dnd-kit/sortable';
 import { restrictToVerticalAxis, restrictToParentElement } from '@dnd-kit/modifiers';
-import { injectGraphql } from 'lib/Injector';
```
emteknetnz: The is unrelated refactoring to remove an unused import

client/src/components/LinkField/LinkField.js
```diff
@@ -38,11 +37,11 @@ const section = 'SilverStripe\\LinkField\\Controllers\\LinkFieldController';
  * ownerID - ID of the owner DataObject
  * ownerClass - class name of the owner DataObject
  * ownerRelation - name of the relation on the owner DataObject
+ * inHistoryViewer - if the field is being viewed in the context of the history viewer
  */
 const LinkField = ({
   value = null,
-  onChange,
-  onNonPublishedVersionedState,
```
emteknetnz: The is unrelated refactoring to remove a non-existent prop

client/src/components/LinkField/LinkField.js
```diff
@@ -94,16 +95,16 @@ const LinkField = ({
...
+          setLoadingError(true);
+        })
+        .finally(() => {
           setLoading(false);
           // isSorting is set to true on drag start and only set to false here to prevent
           // the loading indicator for flickering
           setIsSorting(false);
         })
-        .catch(() => {
-          actions.toasts.error(i18n._t('LinkField.FAILED_TO_LOAD_LINKS', 'Failed to load links'))
```
emteknetnz, line 165: Need to not show a red toast here because this will pop up when viewing old versions of an owner where the link has been deleted.  I've changed this to show an inline message when the link fails to load.

client/src/components/LinkField/LinkField.js
```diff
@@ -350,7 +362,7 @@ const LinkField = ({
 
 LinkField.propTypes = {
   value: PropTypes.oneOfType([PropTypes.arrayOf(PropTypes.number), PropTypes.number]),
-  onChange: PropTypes.func.isRequired,
+  onChange: PropTypes.func,
```
emteknetnz, line 495: This should never have been a required field as it only makes sense in an entwine context

client/src/containers/LinkModalContainer.js
```diff
@@ -23,6 +23,7 @@ const LinkModalContainer = ({ types, typeKey, linkID = 0, isOpen, onSuccess, onC
   // which will causes bugs with validation
   if (!LinkModal) {
     setLinkModal(() => loadComponent(`LinkModal.${handlerName}`));
+    return;
```
emteknetnz, line 26: This should is to fix a console error where the state hasn't yet been updated from the call the setLinkModal, meaning that the `<LinkModal>` jsx object doesn't yet exist at time of render because LinkModal hasn't yet been set. Setting the state via this hook will trigger a rerender in which time `<LinkModal>` will not exist

client/lang/en.js
```diff
@@ -21,7 +21,7 @@ if (typeof(ss) === 'undefined' || typeof(ss.i18n) === 'undefined') {
     "LinkField.LINK_MODIFIED_TITLE": "Link has unpublished changes",
     "LinkField.LINK_MODIFIED_LABEL": "Modified",
     "LinkField.CANNOT_CREATE_LINK": "Cannot create link",
-    "LinkField.FAILED_TO_LOAD_LINKS": "Failed to load links",
+    "LinkField.FAILED_TO_LOAD_LINKS": "Failed to load link(s)",
```
emteknetnz, line 25: This is a minor improvement to the message as it's used for both LinkField and for MultiLinkField

client/src/components/LinkField/LinkField.js
```diff
@@ -38,11 +37,11 @@ const section = 'SilverStripe\\LinkField\\Controllers\\LinkFieldController';
  * ownerID - ID of the owner DataObject
  * ownerClass - class name of the owner DataObject
  * ownerRelation - name of the relation on the owner DataObject
+ * inHistoryViewer - if the field is being viewed in the context of the history viewer
  */
 const LinkField = ({
   value = null,
-  onChange,
-  onNonPublishedVersionedState,
+  onChange = () => {},
```
emteknetnz, line 52: This is a default value that's being set now that it's no longer a required prop

client/src/components/LinkField/tests/LinkField-test.js
```diff
@@ -66,16 +76,21 @@ test('LinkField will render loading indicator if ownerID is not 0', async () =>
   expect(container.querySelectorAll('.link-picker')).toHaveLength(1);
 });
 
-test('LinkField will render link-picker if ownerID is not 0 and has finished loading', async () => {
+test('LinkField will render link-picker if ownerID is not 0 and isMulti and has finished loading', async () => {
```
emteknetnz, line 184: This jest test was technically wrong before - it shouldn't render a `.link-picker` element if it's non-isMulti and ownerID is not 0 - it should only render that if it's isMulti

GuySartorelli (changes requested):
Still need to fulfil this acceptance critera (or deem it too hard for this card and open a separate card for them):

## Archived Admin uses the same behaviour as Versionned Admin when displaying link.

If I click on an archived page from archive admin, I see:
image

Note that there was a link in both the "Has one link" and "Has many links one" field, but not the "Has many links two" field.

## LinkField on versioned history view correctly identify if a Link was attached to the page, but the Link was deleted.

This doesn't seem to work.
1. Create a link on a page
1. Publish the page
1. Archive the link
1. Make a change to the page and publish it again
1. View the version of the page which did have the link

For me it shows "Failed to load link(s)"

## Issue with archived history
When I archive a page (which has links on it), then restore the page from archived, then in the history tab I try to view the archive version, I get an infinite loading component.

emteknetnz:
> Archived Admin uses the same behaviour as Versionned Admin when displaying link.

ArchivedAdmin support adds a level of complexity over the regular history viewer that I don't think is worth it. We need to support multiple permutations of owner and links archived and not-archived. The LinkFieldController which handles all the data isn't aware of this. We also need to add support for knowing if we're even in an archived admin which is very messy because of the special treatment given to SiteTree vs other dataobjects.  

> LinkField on versioned history view correctly identify if a Link was attached to the page, but the Link was deleted.

This relates to the point above.  Because the link was deleted then when attempting to fetch the link via LinkFieldController it will return a 404.  We could change the message from "Failed to load link(s)" to "Link was deleted" ... but then again that's an assumption because it may be 404'ing for a different reason. I'd say it's such a minor point we should just leave as is.

> Issue with archived history

I was unable to replicate this issue

src/Form/LinkField.php
```diff
@@ -4,6 +4,7 @@
 
 use SilverStripe\LinkField\Models\Link;
 use SilverStripe\Forms\HasOneRelationFieldInterface;
+use SilverStripe\VersionedAdmin\Controllers\HistoryViewerController;
```
ChloeSartorelli: Adding this but it's not used?

src/Form/MultiLinkField.php
```diff
@@ -46,6 +47,11 @@ protected function getDefaultAttributes(): array
         return $attributes;
     }
 
+    private function getInHistoryViewer(): bool
+    {
+        return is_a($this->getForm()->getController(), HistoryViewerController::class);
+    }
```
ChloeSartorelli: You have this in the abstract field, I assume you just forgot to remove from here?

GuySartorelli (changes requested):
LGTM, definitely an improvement over how it was before. Just a bit of tidy up.

src/Form/LinkField.php
emteknetnz: Updated

src/Form/MultiLinkField.php
emteknetnz: Updated

src/Form/MultiLinkField.php
```diff
@@ -6,6 +6,7 @@
 use SilverStripe\ORM\DataObject;
 use SilverStripe\ORM\Relation;
 use SilverStripe\ORM\SS_List;
+use SilverStripe\VersionedAdmin\Controllers\HistoryViewerController;
```
ChloeSartorelli:
```suggestion
```

emteknetnz: Updated

GuySartorelli (approved):
LGTM

Issue #213, opened 2024-02-08
Deleting a link does not remove it from live table


emteknetnz:
When a link is deleted, it remains on the live table.

This is because delete() is used instead of doArchive().  The docblock on delete() is wrong in saying that it removes it from both tables

## PRs
- https://github.com/silverstripe/silverstripe-linkfield/pull/214

GuySartorelli:
PR merged

PR #214, opened 2024-02-08
FIX Archive record from live table on delete

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-linkfield/issues/213

### DIFF
--- a/src/Controllers/LinkFieldController.php
+++ b/src/Controllers/LinkFieldController.php
@@ -22,6 +22,8 @@ use SilverStripe\ORM\DataList;
 use SilverStripe\ORM\DataObject;
 use SilverStripe\LinkField\Form\LinkField;
 use SilverStripe\LinkField\Form\MultiLinkField;
+use SilverStripe\ORM\Queries\SQLUpdate;
+use SilverStripe\Versioned\Versioned;
 
 class LinkFieldController extends LeftAndMain
 {
@@ -136,15 +138,10 @@ class LinkFieldController extends LeftAndMain
         if (!SecurityToken::inst()->checkRequest($this->getRequest())) {
             $this->jsonError(400);
         }
-        // delete() will also delete any published version immediately
-        $link->delete();
-        // Update owner object if this Link is on a has_one relation on the owner
-        $owner = $this->getOwnerFromRequest();
-        $ownerRelation = $this->getOwnerRelationFromRequest();
-        $hasOne = Injector::inst()->get($owner->ClassName)->hasOne();
-        if (array_key_exists($ownerRelation, $hasOne) && $owner->canEdit()) {
-            $owner->$ownerRelation = null;
-            $owner->write();
+        if ($link->hasExtension(Versioned::class)) {
+            $link->doArchive();
+        } else {
+            $link->delete();
         }
         // Send response
         return $this->jsonSuccess(204);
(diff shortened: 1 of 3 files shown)


tests/php/Controllers/LinkFieldControllerTest.php
```diff
@@ -500,7 +513,6 @@ public function testLinkDelete(
             $owner = $this->getFixtureLinkOwner();
             $this->assertSame(0, $owner->LinkID);
         }
-        $this->assertTrue(true);
```
emteknetnz, line 503: This is unrelated refactoring to remove a pointless assertion

src/Controllers/LinkFieldController.php
```diff
@@ -136,15 +137,28 @@ public function linkDelete(): HTTPResponse
...
+            // Update draft owner table
             $owner->$ownerRelation = null;
             $owner->write();
+            // Update live owner table if owner is versioned
+            if ($owner->hasExtension(Versioned::class)) {
+                Versioned::withVersionedMode(function () use ($ownerRelation) {
+                    Versioned::set_stage(Versioned::LIVE);
+                    $liveOwner = $this->getOwnerFromRequest();
+                    $liveOwner->$ownerRelation = null;
+                    $liveOwner->write();
```
emteknetnz, line 159: I'm not sure if we should use write() here, or should just do something closer to raw sql that bypasses the ORM

src/Controllers/LinkFieldController.php
```diff
@@ -136,15 +137,28 @@ public function linkDelete(): HTTPResponse
...
+                });
+            }
```
ChloeSartorelli: This incorrectly published changes which are in draft.
To reproduce:
1. Create a link
2. Publish the page
3. Make some content change on the page and save (but don't publish)
4. Archive the link

Reloading the page will show a published state for the page (i.e. publish button is not green).
Visiting the history will show the published state has the draft changes in it.
HOWEVER! Viewing the page in the front-end shows the old published content (not the draft content that got sort-of-maybe-published?)

Not sure if that's a bug with versioned or if this is just being done wrong or what, but something's not right.

GuySartorelli (changes requested):

src/Controllers/LinkFieldController.php
emteknetnz: Updated to use a SQLUpdate instead

src/Controllers/LinkFieldController.php
```diff
@@ -136,15 +138,29 @@ public function linkDelete(): HTTPResponse
...
             $owner->write();
+            // Update live owner table if owner is versioned
+            if ($owner->hasExtension(Versioned::class)) {
+                $draftTable = $owner->getSchema()->tableForField($owner::class, $ownerRelation . 'ID');
+                $liveTable = $draftTable . '_Live';
+                SQLUpdate::create()
+                    ->setTable($liveTable)
+                    ->assign('"' . $ownerRelation . 'ID"', 0)
+                    ->addWhere(['ID' => $owner->ID])
+                    ->execute();
```
ChloeSartorelli: This feels a bit hacky.
It's also not consistent with what happens when you archive related records in other situations. I don't think we should be doing this part.

The ORM already knows how to handle the scenario where a related record doesn't exist anymore. We don't need to try to handle that here.

If the old ID is causing errors in the UI, we need to instead swallow those errors (an outdated ID is fine, we just treat is as though the ID wasn't there in literally every other scenario in the CMS)

ChloeSartorelli: Should have realised this last time, sorry about that.

GuySartorelli (changes requested):

GuySartorelli (approved):
Looks great, works well locally.

There's some bad toasts when looking in the history tab, but that can be handled in https://github.com/silverstripe/silverstripe-linkfield/issues/10

Issue #78, opened 2023-09-14
Standardise module


emteknetnz:
Rough list of things we could do to this module to get it in line with other Silverstripe modules

Note will release an new major so can make API breaks:

Note there's a second look details arou

# Task list

## Architecture
- See https://github.com/silverstripe/silverstripe-linkfield/issues/81

## Code quality
- [x] Add strong typing
- [x] Look at unit tests coverage and add more if it makes sense to
- [x] Consider turning the two LinkField-related traits into an abstract class instead, which `LinkField` and `MultiLinkField` can be subclasses of
  - [x] Look at other code duplication that can also be moved into that abstract class
- [x] Delete unused code (e.g. `client/src/LinkModal/FileLinkModal.js`)

## Maintenance
- [x] Rename .github/workflows/main.yml to .github/workflows/ci.yml
- [x] Run module standardiser, will add things like dispatch-ci.yml and keepalive.yml
- [x] Update module standardiser to remove other old meta files e.g. .codeclimate.yml etc
- [x] Update readme badges, get them consistent with other modules
- [x] Fix typos e.g. LinkTypeResolver.php "provdied"
- [x] Follow instruction for [creating new repo](https://silverstripe.atlassian.net/wiki/spaces/CMS/pages/3030679981/Setting+up+a+new+GitHub+repository).
- [x] Add to transifex
- [x] Add to kitchen sink

## PRs
- https://github.com/silverstripe/silverstripe-linkfield/pull/100
- https://github.com/silverstripe/silverstripe-admin/pull/1678
- https://github.com/silverstripe/silverstripe-linkfield/pull/216
- https://github.com/silverstripe/module-standardiser/pull/38
- https://github.com/silverstripe/supported-modules/pull/27
- https://github.com/silverstripe/recipe-kitchen-sink/pull/48
- https://github.com/silverstripe/silverstripe-linkfield/pull/223

maxime-rainville:
Just a bit of extra context.

## DBLink
The DBLink implementation was something I put in there because @chillu was concerned adding tables for each link type would clog the database. I for one never was a big fan of this.

I think the fact that most people naturally went for the Link DataObject shows that most dev prefer this approach because it's closer to what they are used to.

If we ditch DBLink there's a lot of extra simplication we can put in place. The following classes can be ditch altogether:
- [LinkTypeResolver](https://github.com/silverstripe/silverstripe-linkfield/blob/3/src/GraphQL/LinkTypeResolver.php)
[...]

PR #216, opened 2024-02-12
ENH Refactor: Improve code quality, remove unused code, general tidy up.

GuySartorelli:
Requires https://github.com/silverstripe/silverstripe-admin/pull/1678 to be merged for i18n stuff

Note that I have made changes in discrete commits so it's easier to review each _type_ of change by just looking at each commit individually. Commits can be squashed if you want, though.

## Issue
- https://github.com/silverstripe/silverstripe-linkfield/issues/78


_config/config.yml
```diff
@@ -2,8 +2,14 @@
...
 SilverStripe\Admin\LeftAndMain:
-  extensions:
-    - SilverStripe\LinkField\Extensions\LeftAndMainExtension
+  extra_requirements_javascript:
+    'silverstripe/linkfield:client/dist/js/bundle.js':
+      defer: true
+  extra_requirements_i18n:
+    - 'silverstripe/linkfield:client/lang'
+  extra_requirements_css:
+    - 'silverstripe/linkfield:client/dist/styles/bundle.css'
```
ChloeSartorelli, line 11: Copied from the `LeftAndMainExtension` - no sense having a whole extension for this. I've also raised [a card](https://github.com/silverstripe/.github/issues/194) to do this for other modules.

client/src/components/LinkModal/FileLinkModal.js
ChloeSartorelli, line 1: This wasn't being used and probably doesn't even work with the updated code.

src/Form/AbstractLinkField.php
ChloeSartorelli, line 1: This replaces the traits, and also includes some code that was duplicated across the two linkfield implementations.

src/Models/EmailLink.php
```diff
@@ -20,9 +20,6 @@ class EmailLink extends Link
         'Email' => 'Varchar(255)',
     ];
 
-    /**
-     * Set the priority of this link type in the CMS menu
-     */
     private static int $menu_priority = 30;
```
ChloeSartorelli, line 23: I updated the PHPDoc for this config property in `Link` - rather than trying to keep these in sync, I've opted to just remove the duplicates. API docs are smart enough to carry the description through, and for custom link types you'll probably be looking at `Link` rather than one of the subclasses anyway.

src/Models/Link.php
```diff
@@ -6,10 +6,7 @@
 use ReflectionException;
 use SilverStripe\Core\ClassInfo;
 use SilverStripe\Core\Injector\Injector;
-use SilverStripe\Forms\CompositeValidator;
-use SilverStripe\Forms\DropdownField;
 use SilverStripe\Forms\FieldList;
-use SilverStripe\Forms\RequiredFields;
```
ChloeSartorelli, line 12: Here and anywhere else I've removed `use` statements, it's cause they weren't being used.

src/Models/Link.php
```diff
@@ -55,14 +52,8 @@ class Link extends DataObject
     ];
 
     /**
-     * In-memory only property used to change link type
-     * This case is relevant for CMS edit form which doesn't use React driven UI
-     * This is a workaround as changing the ClassName directly is not fully supported in the GridField admin
-     */
-    private ?string $linkType = null;
```
ChloeSartorelli, line 62: This was used to let you pick a link type if creating a link in a gridfield. This isn't something we want to support, and if people _really_ want to do it they should use `GridFieldAddNewMultiClass` from [`symbiote/silverstripe-gridfieldextensions`](https://github.com/symbiote/silverstripe-gridfieldextensions)

Most (if not all) of the other code removed in this class is related to this feature.

emteknetnz (approved):
Tested locally, all seems to be working as expected

Issue #208, opened 2024-02-07
Use to toMap() in Link::jsonSerialize() could lead to data disclosure


emteknetnz:
The following code exists in [Link::jsonSerialize()](https://github.com/silverstripe/silverstripe-linkfield/blob/4/src/Models/Link.php#L273)

```php
        // TODO: this could lead to data disclosure - we should only return the fields that are actually needed
        $data = $this->toMap();
```

The code is called as part of `LinkFieldController::linkData()`

Data disclosure could happen when custom links or extensions to existing links add data columns that are not supposed to be shown in the CMS

Instead of using the 'include everything' approach, we should have an explicit list of fields returned.  The list should be extensible.

## Acceptance criteria
- Review the usage of `Link::jsonSerialize()` and confirm what its used for and if it's returning fields that aren't needed.
- Re-factor `jsonSerialize` and/or related method to only returned needed fields.
- If there's an inherent residual security risk for custom link types, that risk is called out explicitly in the documentation.
- TODO comment is removed from codebase.
- Consider adding an extension hooks if there's a probably use case for it.

## Affected code snippet
https://github.com/silverstripe/silverstripe-linkfield/blob/c75615f93d5fb3c899e2b9ee40e8528614e0a644/src/Models/Link.php#L206-L224

## Extra context
- Being able to JSON serialise Links used to be a big deal because all the data was being passed via JSON string. It's a much smaller concern now.
- Link used to extend the [`JsonSerializable`](https://www.php.net/manual/en/jsonserializable.jsonserialize.php) interface. `JsonSerializable::jsonSerialize()` allows you to customise how an object gets represented when `json_encode` is called on it.
- I would strongly suggest either:
  - Reimplementing the `JsonSerializable` interface if there's still a need to be able to json serialise Links ... and using `json_encode` instead of directly calling `jsonSerialize` directly.
  - Rename `jsonSerialize` to a more descriptive name if it's not actually serializing the Link.
  
## PRs
- https://github.com/silverstripe/silverstripe-linkfield/pull/233

PR #233, opened 2024-02-19
ENH Avoid potential data disclosure

GuySartorelli:
## Issue
- https://github.com/silverstripe/silverstripe-linkfield/issues/208


client/src/components/LinkField/LinkField.js
```diff
@@ -353,20 +353,20 @@ const LinkField = ({
...
+        description={linkData.description}
+        versionState={linkData.versionState}
         typeTitle={type.title || ''}
         typeIcon={type.icon}
         onDelete={handleDelete}
         onClick={() => { setEditingID(linkID); }}
         onButtonKeyDownEdit={() => setIsKeyboardEditing(true)}
         onUnpublishedVersionedState={handleUnpublishedVersionedState}
-        canDelete={data[linkID]?.canDelete ? true : false}
+        canDelete={linkData.canDelete ? true : false}
```
ChloeSartorelli, line 369: Unrelated refactoring - this variable is explicitly created above so a) let's use it and b) no need for null safety chaining operator.

src/Controllers/LinkFieldController.php
```diff
@@ -118,11 +118,7 @@ private function getLinkData(Link $link): array
         if (!$link->canView()) {
             $this->jsonError(403);
         }
-        $data = $link->jsonSerialize();
-        $data['canDelete'] = $link->canDelete();
-        $data['description'] = $link->getDescription();
-        $data['versionState'] = $link->getVersionedState();
-        return $data;
+        return $link->getData();
```
ChloeSartorelli, line 121: Moved these into the method on `Link` since the only time the method is called is here. Makes it easier to maintain and see the full list of what's passed through to the react component from PHP in one place.

src/Models/Link.php
```diff
@@ -201,33 +201,30 @@ public function onBeforeWrite(): void
...
-
-        $data['Title'] = $this->getTitle();
-
-        return $data;
+        return [
+            'title' => $this->getTitle(),
+            'description' => $this->getDescription(),
+            'canDelete' => $this->canDelete(),
+            'versionState' => $this->getVersionedState(),
+            'typeKey' => $typeKey,
```
ChloeSartorelli, line 219: Explicitly only these are used in the component.
Even creating super-duper unique link subclasses won't require adding more data here - the modal is handled separately, this is just for display before opening the modal.

There's no way to easily customise the react component, so adding an extension hook to update this data is overkill.

src/Models/Link.php
```diff
@@ -201,33 +201,30 @@ public function onBeforeWrite(): void
...
     }
 
     /**
-     * Return a rendered version of this form.
+     * Return a rendered version of this link.
      *
-     * This is returned when you access a form as $FormObject rather
-     * than <% with FormObject %>
+     * This is returned when you access a link as $LinkRelation or $Me rather
+     * than <% with LinkRelation %>
```
ChloeSartorelli, line 227: Unrelated refactoring - a link is not a form, this was probably just copy pasted from elsewhere originally.

client/src/components/LinkField/LinkField.js
```diff
@@ -353,20 +353,20 @@ const LinkField = ({
...
       }
-      const type = types.hasOwnProperty(data[linkID]?.typeKey) ? types[data[linkID]?.typeKey] : {};
+      const type = types.hasOwnProperty(linkData.typeKey) ? types[linkData.typeKey] : {};
       links.push(<LinkPickerTitle
         key={linkID}
         id={linkID}
-        title={data[linkID]?.Title}
-        description={data[linkID]?.description}
-        versionState={data[linkID]?.versionState}
+        title={linkData.title}
```
ChloeSartorelli, line 360: Note upper case `Title` is now lower case `title` - all of the data is now using lowerCamelCase key names, as per javascript tradition. It was a mix before.

emteknetnz (approved):

emteknetnz:
Tested locally, works good

Issue #235, opened 2024-02-25
update docs links in readme


GuySartorelli:
The readme has a link to the user guide, but the text for the link implies it's the dev docs.
We should have clear links to both the user guide **and** the dev docs.

## PRs
- https://github.com/silverstripe/silverstripe-linkfield/pull/236

PR #236, opened 2024-02-25
DOC Fix docs links in readme

GuySartorelli:
Should link to both dev docs and user help.

Links obviously won't work until those docs are released in April.

## Issue
- https://github.com/silverstripe/silverstripe-linkfield/issues/235


emteknetnz (approved):
MOG

PR #241, opened 2024-03-01
ENH Pre-render the LinkField in entwine

maxime-rainville:
This PR is making the pre-rendered UI more pleasing by pre-rendering somethings.

## Before
[no-pre-load-linkfield-2024-03-01_21.55.02.webm](https://github.com/silverstripe/silverstripe-linkfield/assets/1168676/4b8db38f-379c-4cb8-9ba4-6977eba0af37)

### Before React boots
Screenshot from 2024-03-01 21-57-31

### Before LinkField gets its data back
Screenshot from 2024-03-01 22-00-00

## After
[pre-load-linkfield-2024-03-01_21.53.59.webm](https://github.com/silverstripe/silverstripe-linkfield/assets/1168676/db0ad13d-1c49-4666-9a76-1c0ec0da49a0)

### Before React boots
Screenshot from 2024-03-01 22-02-10

### Before LinkField gets its data back
image

## Parent issue
- https://github.com/silverstripe/silverstripe-linkfield/pull/242

## Depends on
- https://github.com/silverstripe/silverstripe-admin/pull/1705


client/src/components/LinkField/LinkField.js
```diff
@@ -348,12 +348,12 @@ const LinkField = ({
     const links = [];
     for (let i = 0; i < linkIDs.length; i++) {
       const linkID = linkIDs[i];
-      // Only render items we have data for
-      const linkData = data[linkID];
-      if (!linkData) {
+      // Render dataless item to provide a good loading experience, except if we have single link field
+      const linkData = data[linkID] || {};
+      if (!linkData && !isMulti) {
         continue;
```
maxime-rainville, line 359: By not skipping the link entries for which we don't have data yet, we'll render a list of placeholder LinkPickerTitle. This will avoids having a big jump on the page once the data finally gets back to us.

client/src/components/LinkField/LinkField.js
```diff
@@ -446,7 +446,7 @@ const LinkField = ({
 
   const saveRecordFirst = !loadingError && ownerID === 0;
   const renderLoadingError = loadingError;
-  const renderPicker = !loadingError && !inHistoryViewer && !saveRecordFirst && (isMulti || Object.keys(data).length === 0);
```
maxime-rainville, line 449: This was needed because a disabled LinkField with pe-existing data would briefly render a picker and a place holder link.

With this change, it only renders the picker.

src/Form/MultiLinkField.php
```diff
@@ -132,4 +133,9 @@ private function loadFrom(DataObject $record): void
         $value = array_values($relation->getIDList() ?? []);
         parent::setValue($value);
     }
+
+    public function LinkIDs(): ArrayList
+    {
+        return ArrayList::create($this->dataValue());
```
maxime-rainville, line 139: This wraps the IDs in an ArrayList so the SS template can loop over them.

templates/SilverStripe/LinkField/Form/LinkField_Spinner.ss
maxime-rainville, line 1: Maybe this should be in Admin instead. [CMSLoadingScreen.ss](https://github.com/silverstripe/silverstripe-admin/blob/2/templates/SilverStripe/Admin/Includes/CMSLoadingScreen.ss) contains the relevant markup, but it also contains other things that get in the way.

GuySartorelli (changes requested):
Are the new templates always used for rendering those elements, or does react render them separately?
If the latter, this gives the false impression that you can change how the linkfield gets rendered by changing templates.

maxime-rainville:
Enwtine only.

Most of these templates were pre-existing, so you could easily have been confused into thinking that you could override them. If you did try to customise them, you would quickly notice that your changes a being nuked once React boots.

I can rename the templates and/or add extra comments to make it clear this is an Entwine only construct.

GuySartorelli:
> Enwtine only.

Linkfield is a react component, so that doesn't really make any sense

> Most of these templates were pre-existing, so you could easily have been confused into thinking that you could override them. If you did try to customise them, you would quickly notice that your changes a being nuked once React boots.

The existing ones **CAN** be overridden, but notably are used to bootstrap the react component.

The new templates mirror something that is rendered by react - so changing these templates will **only** change the display of a linkfield before any react is bootstrapped for the field if I'm not mistaken.

At the very very least these templates need comments at the top indicating that they are only for use in rendering the field before react has bootstrapped and will not affect the display of the rendered react component.

maxime-rainville:
Behat build is broken. I thought it was intermittent, but it seems to be a legit failure.

I'll add comments to the template while I'm at it.

tests/behat/features/accessibility-linkfield.feature
```diff
@@ -96,6 +96,7 @@ Feature: Accessibility Tests
 
     Then I press the "Tab" key globally
     And I press the "Tab" key globally
+    And I press the "Tab" key globally
```
maxime-rainville, line 99: I had to add an extra Tab here because for some reason the focus is not returned to the LinkField after the Modal is saved.

The expected course of action is:
- You close the Modal, the the LinkField has focus.
- The First Tab shift the focus to the Archive button
- The Second Tab shift the focus to the next LinkField.

With the current PR:
- You close the Modal, the the LinkField doesn't have focus.
- The First Tab shift the focus to the LinkField
- The Second Tab shift the focus to the Archive button
- The Third Tab shift the focus to the next LinkField.

After screaming at the thing for a couple hours, I concluded that the initial implementation doesn't really work most of the time any way: 
- Some fields won't return the focus to the right place.
- Some field will work on the first Modal edition, but not on the second.

I got a linkfield accessibility follow up card here: https://github.com/silverstripe/silverstripe-linkfield/issues/245

emteknetnz: > You close the Modal, the the LinkField doesn't have focus.

Sound like the PR has created a regression then? Before this PR `You close the Modal, the the LinkField has focus.` worked correctly with keyboard navigation (though not with mouse), though this behat test is using keyboard navigation so the fact you need to tab three times is wrong?

> Some fields won't return the focus to the right place.
> Some field will work on the first Modal edition, but not on the second.

Could you provide some details please

templates/SilverStripe/LinkField/Form/LinkField_Spinner.ss
```diff
@@ -0,0 +1,26 @@
+  <div class="link-field__loading">
```
emteknetnz: Why are we using templates here?  This won't work in an react context e.g. elemental

There's a spinner react component available, seems like that should be used instead

client/src/components/LinkField/tests/LinkField-test.js
```diff
@@ -89,6 +89,7 @@ test('LinkField will render disabled state if disabled is true', async () => {
   const { container } = render(<LinkField {...makeProps({
     ownerID: 1,
     disabled: true,
+    value: undefined
```
emteknetnz, line 110: Why is this set to undefined instead of the default 123? A disabled LinkField could still have a value?

emteknetnz (changes requested):
We shouldn't be adding templates to a react component

templates/SilverStripe/LinkField/Form/LinkField_Spinner.ss
maxime-rainville: The entire point of the PR is to render something while we wait for React to boot.

Elemental won't have that problem, because it's already powered by react.

tests/behat/features/accessibility-linkfield.feature
maxime-rainville: Without this PR, the only time the focus works is:
- When creating a brand new link
- When exiting a modal to an empty link field

It does NOT work:
- When updating an existing Link
- When exiting the modal for a link that already exists
- When archiving a link

From where I stand, the focus is already busted anyway and needs to be redone. So I don't think we should hold this PR on this.

client/src/components/LinkField/tests/LinkField-test.js
maxime-rainville: The original version of the test was incorrect. A disabled LinkField with a value should not be rendering `Cannot create link` message. It should be rendering the Link in a disabled state.

I added a new test case. Now, we test:
- Disabled LinkField with link
- Disabled LinkField without link

I've removed the `doResolve` call from the "Disabled without" scenario because there's no actual AJAX call made in this instance

client/src/components/LinkField/tests/LinkField-test.js
```diff
@@ -86,12 +86,30 @@ test('LinkField returns list of links if they exist', async () => {
...
+  />);
+  await doResolve({ json: () => ({
+    123: {
+      title: 'Page title',
+      typeKey: 'mylink',
+    }
+  }) });
+  await screen.findByText('Page title');
+  const linkPicker = container.querySelectorAll('#link-picker__link-123');
+  expect(linkPicker[0]).toHaveAttribute('aria-disabled', 'false');
```
maxime-rainville: There's an `aria-disabled` attribute on the Disabled linkfield ... but its value is false. I'm not sure if that's actually what we want to do here.

I've added like this for now. We can revisit in the accessibility card.

maxime-rainville (commented):
I've addressed the feedback.

On the broader point of "adding templates to a react component", providing placeholder markup while waiting for some long running process to complete is a standard practice in most modern JS based application.

For example, here's what ZenHub does.
image

It provides a nicer UI for the user, avoid things jumping around on the page once the long process has completed and reassures the user that the site hasn't crash on them.

I appreciate we have not historically done this. Now is as good a time to start as any.

client/src/components/LinkField/tests/LinkField-test.js
emteknetnz:
```suggestion
```
Adding it here make it look like this is the correct behaviour, even though we know it's wrong. If we're explicitly adding aria-disabled on the other card then we'll naturally add a unit test for it

tests/behat/features/accessibility-linkfield.feature
emteknetnz: > When updating an existing Link
> When exiting the modal for a link that already exists

I have replicated locally, though it's a sporadic failure. Seems like a timing issue. 

> When archiving a link

This was never accounted for

emteknetnz (changes requested):
There's little orange 'modified' badges showing before it's loaded as part of the 'loading animation' which shouldn't be there

This video showing me refreshing the page on my local

[linkfield-fout-after-2024-03-07_15.25.41.webm](https://github.com/silverstripe/silverstripe-linkfield/assets/4809037/de085926-a919-4857-9d83-490b4c56ec2a)

I can't help thinking we need to look at the spinner styling, the dark background on it make for a pretty jarring experience on my local where there's lots of "flashing" because the background is rendered for so little time. There already was some before this PR, though it's quite a lot worse afterwards

templates/SilverStripe/LinkField/Form/LinkField_Spinner.ss
ChloeSartorelli, line 1: This template is missing the comment about how it's only for pre-react rendering, and will be discarded.

maxime-rainville: You're meant to be on leave, sir!

client/src/components/LinkField/tests/LinkField-test.js
maxime-rainville: Removed the test for false.

client/src/constants/versionStates.js
```diff
@@ -0,0 +1,9 @@
+const versionStates = {
```
maxime-rainville, line 1: Repeating the same strings over and over again seemed wrong and likely to lead to confusion. So I moved them to a constant.

client/src/components/LinkPicker/LinkPickerTitle.js
```diff
@@ -198,4 +201,8 @@ LinkPickerTitle.propTypes = {
   buttonRef: PropTypes.object.isRequired,
 };
 
+LinkPickerTitle.defaultProps = {
+  versionState: versionStates.unversioned,
```
maxime-rainville, line 205: This change addresses the problem where the place holder LinkPickerTitle would get a draft state icon.

composer.json
```diff
@@ -17,7 +17,8 @@
         "php": "^8.1",
         "silverstripe/framework": "^5.2",
         "silverstripe/cms": "^5",
-        "silverstripe/versioned": "^2"
+        "silverstripe/versioned": "^2",
+        "silverstripe/admin": "^2.2"
```
maxime-rainville, line 21: We have a dependency on admin 2.2 because of the new spinner template.

templates/SilverStripe/LinkField/Form/LinkField_Spinner.ss
maxime-rainville: Added the comments + moved the spinner to a dedicated template in admin so we don't need to repeat this over and over.

See https://github.com/silverstripe/silverstripe-admin/pull/1705

maxime-rainville (commented):
Added the missing comment and cleaned the version state icon.

FYI this PR now depends on https://github.com/silverstripe/silverstripe-admin/pull/1705

client/src/constants/versionStates.js
emteknetnz: Could you please refrain from doing random refactoring half-way through a peer review. It adds unnecessary burden at the end of the person doing the peer review.

emteknetnz (approved):
Tested locally

emteknetnz (changes requested):
Behat failures in CI

maxime-rainville:
Merging the admin PR fixed the behat build.

emteknetnz (approved):

Issue #243, opened 2024-03-03
LinkField needs fluent's gridfield removed so modal renders


emteknetnz:
Fluent will add a tab with a gridfield to linkfield, however because there is no react component for gridfield, you'll get a hard error on the frontend and the link modal will not work at all

To work around this on the [SPIKE](https://github.com/silverstripe/silverstripe-linkfield/issues/210#issuecomment-1943029385) added a workaround `protected $schemaDataType = self::SCHEMA_DATA_TYPE_STRUCTURAL;`, though this shouldn't be merged as the lack of a gridfield component should hard fail so that developers are prompted to remove gridfields

## Acceptance criteria
- When fluent is installed, the tab with the gridfield is removed so that it does not render in the link modal

## PR
- https://github.com/silverstripe/silverstripe-linkfield/pull/247

GuySartorelli:
PR merged

PR #247, opened 2024-03-07
FIX Remove tabs with only gridfields

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-linkfield/issues/243

### DIFF
--- a/src/Extensions/FluentLinkExtension.php
+++ b/src/Extensions/FluentLinkExtension.php
@@ -1,0 +1,35 @@
+<?php
+
+namespace SilverStripe\LinkField\Extensions;
+
+use SilverStripe\Core\Extension;
+use SilverStripe\Forms\GridField\GridField;
+use SilverStripe\Forms\FieldList;
+use TractorCow\Fluent\Extension\FluentVersionedExtension;
+
+class FluentLinkExtension extends Extension
+{
+    public function updateCMSFields(FieldList $fields)
+    {
+        $this->removeTabsWithFluentGridfields($fields);
+    }
+
+    /**
+     * Remove tabs added by fluent that contain GridFields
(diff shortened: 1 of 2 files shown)


GuySartorelli (changes requested):
We shouldn't remove all gridfields. This is explicitly about removing the gridfield that fluent adds.
If people are adding other gridfields, they should still get hard errors telling them they shouldn't be doing that.

emteknetnz:
Updated

src/Extensions/FluentLinkExtension.php
ChloeSartorelli, line 1: This feels like a massive over-complication... Why not just use `removeByName`?

i.e:
```php
if ($this->getOwner()->hasExtension(TheFluentExtension::class)) {
    $fields->removeByName(['Locales', 'FilteredLocales']);
}
```

emteknetnz: We could be that directly in getCMSFields(), though it's feels kind of icky having fluent specific code in there since fluents an optional module, I like having it in its own conditional extension so things are clearly separate

ChloeSartorelli: Right... I'm not suggesting doing this in `getCMSFields()` though. I'm suggesting doing this here in this extension (hence `getOwner`). I don't understand why the logic is as complex as it is in this extension, when we know (and are even using already) the exact names of the tabs/fields we want to remove. Seems like we should just remove them, unless I'm missing something?

emteknetnz: Oh right I see what you mean. Have updated.

GuySartorelli (approved):
Works well locally, tested with has_one localised link.

Issue #240, opened 2024-03-01
FileLink default text link should be the File Title


maxime-rainville:
### Module version(s) affected

4.x-dev

### Description

File Link should use the File title as their default title. It's currently using the Filename including the path.

![image](https://github.com/silverstripe/silverstripe-linkfield/assets/1168676/9c97f855-de85-4063-a8e4-db88a8d50049)

### How to reproduce

1. Add a link to file.
2. Do not provide Link text.
3. Save your link text

Expected results: the Title of the link matches the title of the file.
Actual: The Filename of the File is the Title of the link

### Possible Solution

Fix this line:
https://github.com/silverstripe/silverstripe-linkfield/blob/b6dcef0a273e43267ed1f608c461414bf7c716af/src/Models/FileLink.php#L53

### Acceptance criteria
- The default title of FileLink is the File Title.

### Validations

- [X] Check that there isn't already an issue that reports the same bug
- [X] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

### PR
- https://github.com/silverstripe/silverstripe-linkfield/pull/248

maxime-rainville:
This should be trivial to fix, but I'm putting it in Refinement first to give people a chance to disagree with me.

GuySartorelli:
PR merged

PR #248, opened 2024-03-11
ENH Use File Title

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-linkfield/issues/240

### DIFF
--- a/src/Models/FileLink.php
+++ b/src/Models/FileLink.php
@@ -74,7 +74,6 @@ class FileLink extends Link
         if (!$file?->exists()) {
             return _t(__CLASS__ . '.MISSING_DEFAULT_TITLE', '(File missing)');
         }
-
-        return (string) $this->getDescription();
+        return $file->Title;
     }
 }
--- a/tests/behat/features/create-edit-linkfield.feature
+++ b/tests/behat/features/create-edit-linkfield.feature
@@ -147,7 +147,7 @@ I want to add links to pages, files, external URLs, email addresses and phone nu
 
     # Fourth link in multi link field
     And I should see "Link to a file" in the "[data-field-id='Form_EditForm_HasManyLinks'] .link-picker__link:nth-of-type(4)" element
-    And I should see "folder1/file1.jpg" in the "[data-field-id='Form_EditForm_HasManyLinks'] .link-picker__link:nth-of-type(4)" element
+    And I should see "File1" in the "[data-field-id='Form_EditForm_HasManyLinks'] .link-picker__link:nth-of-type(4)" element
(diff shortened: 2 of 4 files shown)


tests/behat/features/create-edit-linkfield.feature
```diff
@@ -147,7 +147,7 @@ I want to add links to pages, files, external URLs, email addresses and phone nu
 
     # Fourth link in multi link field
     And I should see "Link to a file" in the "[data-field-id='Form_EditForm_HasManyLinks'] .link-picker__link:nth-of-type(4)" element
-    And I should see "folder1/file1.jpg" in the "[data-field-id='Form_EditForm_HasManyLinks'] .link-picker__link:nth-of-type(4)" element
+    And I should see "File1" in the "[data-field-id='Form_EditForm_HasManyLinks'] .link-picker__link:nth-of-type(4)" element
```
ChloeSartorelli, line 150: I think we should see both, right?
There's the default title which is now "File1", but we should also see the URL "folder1/file1.jpg" after "Link to a file"

image

GuySartorelli (changes requested):

tests/behat/features/create-edit-linkfield.feature
emteknetnz: I don't think that's correct, point of this PR is to not show the file url and to show the file title instead

This is how it looks with this PR with and without link text specified. Note that $file->Title is automatically generated in File::onBeforeWrite(), though in this case I've manually adjusted it

image

ChloeSartorelli: Oh, interesting. I'm not sure that's the intended result? Might need to double check with Max or UX. My understanding is that only the default title was meant to change, not the text after "Link to a file".

src/Models/FileLink.php
```diff
@@ -50,7 +50,7 @@ public function getDescription(): string
         if (!$file->canView()) {
             return _t(__CLASS__ . '.CANNOT_VIEW_FILE', 'Cannot view file');
         }
-        return $file->getFilename() ?? '';
+        return $file->Title ?? '';
```
ChloeSartorelli: I think this is meant to stay as is, and only `getDefaultTitle()` is meant to change.

GuySartorelli (changes requested):

src/Models/FileLink.php
emteknetnz: Ah yup makes sense, updated

tests/behat/features/create-edit-linkfield.feature
emteknetnz: Has been changed

GuySartorelli (approved):
LGTM

Issue #245, opened 2024-03-04
LinkField accessibility


maxime-rainville:
As a content author I want LinkField to adhere to accessibility best practices so that I can use the field regardless of any impairment.

## Acceptance criteria

- Accessibility practices for dndkit are adhere to.
- Focus is consistent between keyboard and mouse usages.
- When a user is done editing a Link in the LinkModal focus is returned to 
  - to the LinkField that was being edited (Single LinkField)
  - to the Link being edited (Multi LinkField)
- When archiving a link, the focus is returned to a suitable element (ask UX team for feedback on this one)
- Link list should be an actual list.
- Code is re-factored to avoid bad React practices like using Intervals or directly setting attribute on DOM elements. - **update** see [comment below](https://github.com/silverstripe/silverstripe-linkfield/issues/245#issuecomment-2024564292)
- Fix is targeted to next minor (October)

## Potential problems

- The focus before and after the Modal closes depends on whether you are using the Keyboard or the Mouse.
- When using the keyboard to save the Modal, the cursor appears to be on the correct spot, but the button doesn't actually have tho focus. So if I hit the `Enter` key, nothing happens.
- Our [sorting library has accessibility recommendation](https://docs.dndkit.com/guides/accessibility). It's not clear that they've been applied across the board.
  - e.g. There's no screen reader announcements.
- There's unusual logic in there around setting interval, manually detecting key presses, directly setting attribute on DOM elements.
- Some of the Aria attribute seem poorly defined.
  - The Div containing the LinkField button has a role of `button`. The button inside there also has a the role of `button`.
  - Disabled LinkFIeld have `aria-disabled="false"` applied to them.
  - `aria-describedby` is empty. It should point to the label.
  - In MultiLinkField, the list of link should be define as an actual list ... either with UL or with the equivalent aria-role.
- The link list is not an actual list.

## Related

We had a previous card for this: https://github.com/silverstripe/silverstripe-linkfield/pull/200

## New issues created
- https://github.com/silverstripe/silverstripe-linkfield/issues/263
- https://github.com/silverstripe/silverstripe-linkfield/issues/265

## PRs
- https://github.com/silverstripe/silverstripe-linkfield/pull/250
- https://github.com/silverstripe/silverstripe-linkfield/pull/260
- https://github.com/silverstripe/silverstripe-linkfield/pull/264
- https://github.com/silverstripe/developer-docs/pull/479

maxime-rainville:
## UX Questions

[...]

PR #250, opened 2024-03-13
ENH Use li for multi links

emteknetnz:

- Changes MultiLinkField to use `<ol><li>` for links - single LinkField continues to use <div>
- Added inline comments to explain why `<LinkModalContainer>` in rendered in two different places


client/src/components/LinkField/LinkField.js
```diff
@@ -483,6 +485,7 @@ const LinkField = ({
           isKeyboardEditing={isKeyboardEditing}
         /> }
       {sortableLinks()}
+      { /* This <LinkModalContainer> is only used for editing EXISTING links */ }
```
emteknetnz, line 488: This isn't a debugging comment, this is useful context because there's 2x `<LinkModalContainer>` and I've labelled what each of them do

client/src/components/LinkField/LinkField.js
```diff
@@ -396,12 +396,14 @@ const LinkField = ({
...
-            {links}
-          </SortableContext>
+          <ul className="link-picker__list">
+            <SortableContext
+              items={linkIDs}
+              strategy={verticalListSortingStrategy}
+            >
+              {links}
+            </SortableContext>
+          </ul>
```
```suggestion
          <ol className="link-picker__list">
            <SortableContext
              items={linkIDs}
              strategy={verticalListSortingStrategy}
            >
              {links}
            </SortableContext>
          </ol>
```
Semantics is important here - and semantically since you can reorder the links and the order is important in the CMS context, `ol` should be used (regardless of the fact that in the frontend they're probably going to end up being in an unordered list)

GuySartorelli (changes requested):

client/src/components/LinkField/LinkField.js
emteknetnz: Updated

GuySartorelli (approved):
LGTM

PR #260, opened 2024-03-21
ENH Improve component focusing

emteknetnz:

This PR:
- Makes focus when using mouse and keyboard consistent. Previously only using the keyboard has special logic for setting the focus state
- Fixes a timing issue with focus when using a keyboard
- Swaps around the tab order of the drag handle so that it is last rather than first


client/src/components/LinkModal/LinkModal.js
```diff
@@ -86,7 +86,6 @@ LinkModal.propTypes = {
   isOpen: PropTypes.bool.isRequired,
   onSuccess: PropTypes.func.isRequired,
   onClosed: PropTypes.func.isRequired,
-  autoFocus: PropTypes.bool,
```
ChloeSartorelli, line 89: This prop is still being deconstructed from props in line 32

GuySartorelli (changes requested):
When I archive a link and then reorder links via the keyboard I get this error in the console:
> Assertion failed: Input argument is not an HTMLInputElement

I don't get that if I reorder with the mouse.

GuySartorelli:
Other than the specific problems above, it works great locally.

client/src/components/LinkModal/LinkModal.js
emteknetnz: I've removed it from line 32, it was unused

emteknetnz:
> When I archive a link and then reorder links via the keyboard I get this error in the console:

I'm unable to replicate this. Could you please double check this is still happening with the latest version of this PR and if so then demo it to me.

GuySartorelli:
Can't replicate that problem anymore, no idea what was causing it.
LGTM but has merge conflicts.

emteknetnz:
Done

GuySartorelli (approved):
LGTM

Issue #258, opened 2024-03-19
New Link get a Sort of 1


maxime-rainville:
I suspect that part of the problem is that new Link, seem to default to having a `Sort` of `1`.

## Repo steps
- Start with a list of 4 links.
- Sort them around so they all have an order set to them.
- Publish your changes
- Create a brand new link.
- New link is added at the end of the list
- Refresh the page

Expected results: New link is at the end of the list.
Actual results: New link is in the second spot.

If you look at the DB, you'll find that the new link has a Sort of 1. It's showing up in the second spot because one other link will have a Sort of 1.

[linkfield-2024-03-18_17.07.39.webm](https://github.com/silverstripe/silverstripe-linkfield/assets/1168676/4188fe0e-1ed6-48bb-9c80-d06fbcbfe495)

## Acceptance criteria
- Newly created links in a Link List are added to the end of the list and have the matching sort value assigned to them.

## Related card
https://github.com/silverstripe/silverstripe-linkfield/issues/254

## PR
- https://github.com/silverstripe/silverstripe-linkfield/pull/262

GuySartorelli:
PR merged

PR #262, opened 2024-03-26
FIX Linkfield sorting issue

sabina-talipova:
## Description
The `self::get()` method returns list of specific Link types, because of this for each link type we reset and begin to increase `Sort` again.

## Parent issue
- https://github.com/silverstripe/silverstripe-linkfield/issues/258

### DIFF
--- a/src/Models/Link.php
+++ b/src/Models/Link.php
@@ -191,7 +191,7 @@ class Link extends DataObject
         // Ensure a Sort value is set and that it's one larger than any other Sort value for
         // this owner relation so that newly created Links on MultiLinkField's are properly sorted
         if (!$this->Sort) {
-            $this->Sort = self::get()->filter([
+            $this->Sort = Link::get()->filter([
                 'OwnerID' => $this->OwnerID,
                 'OwnerRelation' => $this->OwnerRelation,
             ])->max('Sort') + 1;
--- a/tests/php/Models/LinkTest.php
+++ b/tests/php/Models/LinkTest.php
@@ -24,6 +24,7 @@ use SilverStripe\LinkField\Tests\Models\LinkTest\LinkOwner;
 use SilverStripe\LinkField\Services\LinkTypeService;
 use SilverStripe\Core\Injector\Injector;
 use SilverStripe\LinkField\Tests\Controllers\LinkFieldControllerTest\TestPhoneLink;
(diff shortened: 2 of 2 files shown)


tests/php/Models/LinkTest.php
```diff
@@ -479,9 +480,66 @@ public function provideLinkType(): array
...
+            $link->write();
+            array_push($linkIds, $link->ID);
+        }
+
+        return $linkIds;
+    }
+
+    public function testSorting(): void
+    {
+        $linkIDs = $this->createLinks();
```
sabina-talipova, line 536: I added new `createLinks ` method since `@dataProvider` annotation doesn't work as expected (want to be) in this specific case.

src/Models/Link.php
```diff
@@ -191,7 +191,7 @@ public function onBeforeWrite(): void
         // Ensure a Sort value is set and that it's one larger than any other Sort value for
         // this owner relation so that newly created Links on MultiLinkField's are properly sorted
         if (!$this->Sort) {
-            $this->Sort = self::get()->filter([
+            $this->Sort = Link::get()->filter([
```
ChloeSartorelli, line 194: I think you may have found a bug in PHP itself! `self` always refers to the class where the code is declared, so these **should** be identical. But for some reason the late static binding inside the `get()` call is still bound to the specific subclass.

We should probably avoid calling static methods in general if this sort of thing is going to happen.
Nice find.

tests/php/Models/LinkTest.php
ChloeSartorelli: What is is about it that doesn't work? I'm guessing it's the calls to `write()` failing due to no DB existing yet?
This is fine anyway for this specific test.

GuySartorelli (approved):
Works well locally.

tests/php/Models/LinkTest.php
sabina-talipova: My problem was that for each subarray in the DataProvider we update the DB records. I'll assume that "TRUNCATE" is used and all previous entries are deleted. The `yml` fixture is also not entirely suitable for me, since I need an array of specific IDs to control the process.

sabina-talipova: Note: I didn't want to change our `setUp` and `tearDown` logic.

PR #264, opened 2024-03-28
ENH Better use of ARIA attributes

emteknetnz:

The fixes a few of things noted in the issue:
- `The Div containing the LinkField button has a role of button. The button inside there also has a the role of button.`
- `Disabled LinkFIeld have aria-disabled="false" applied to them.`
- A DOM manipulation being done in react

Core issue was that that the dndkit attributes were mistakenly on a div instead of child drag handle which is specifically there for accessibility

Regarding `aria-describedby is empty. It should point to the label.` - The drag handle should now have something like `aria-describedby="DndDescribedBy-0"` which points to a hidden div that dndkit underneath the sortable links which will look like this

```html
<div id="DndDescribedBy-0" style="display: none;">
  To pick up a draggable item, press the space bar.
  While dragging, use the arrow keys to move the item.
  Press space again to drop the item in its new position, or press escape to cancel.
</div>
```


client/src/components/LinkPicker/tests/LinkPickerTitle-test.js
```diff
@@ -183,14 +183,3 @@ test('dnd handler is not displayed if link field is not MultiLinkField', () => {
   </LinkFieldContext.Provider>);
   expect(container.querySelectorAll('.link-picker__drag-handle')).toHaveLength(0);
 });
-
-test('keydown on dnd handler', async () => {
```
emteknetnz, line 186: Removing this test because the aria-pressed attribute is now controlled by dndkit's attributes, not with custom code.  In a jest context this fails though real-life it works fine

client/src/components/LinkPicker/LinkPickerTitle.js
```diff
@@ -116,25 +107,24 @@ const LinkPickerTitle = ({
...
-    {...attributes}
     {...listeners}
     id={idAttr}
+    aria-disabled={readonly || disabled}
   >
     { (isMulti && !readonly && !disabled) && <div className="link-picker__drag-handle"
         tabIndex="0"
         role="button"
-        aria-pressed="false"
+        aria-pressed={attributes['aria-pressed'] || false}
```
emteknetnz: dndkit will set a value of undefined for aria-pressed when its not being sorted which is kind of odd and the attribute won't show at all, so I've given it a default value of false instead.

GuySartorelli:
> The Div containing the LinkField button has a role of button. The button inside there also has a the role of button.

While that has been resolved, we now have the situation where there's nothing indicating for screen readers that pressing enter/space on the link element itself will open the edit modal. That's what the `aria-role="button"` was there for.

client/src/components/LinkPicker/LinkPickerTitle.js
```diff
@@ -116,25 +107,24 @@ const LinkPickerTitle = ({
...
-  // Remove the default tabindex="0" attribute from the sortable element because we're going to manually
-  // add this to the drag handle instead
-  delete attributes.tabIndex;
+
   const idAttr = `link-picker__link-${id}`;
   return <div
     className={className}
     ref={setNodeRef}
     style={style}
-    {...attributes}
```
ChloeSartorelli, line 124: It seems a bit strange to not just be passing all the attributes in - are we 1000% sure we aren't accidentally ignoring attributes we want by doing this? What if the attributes change in a future release of the drag and drop library and we are then omitting something that it needs?

GuySartorelli (changes requested):
Seems to work well locally but I do have one concern below and one above.

emteknetnz:
> While that has been resolved, we now have the situation where there's nothing indicating for screen readers that pressing enter/space on the link element itself will open the edit modal. That's what the aria-role="button" was there for.

We don't to define `role="button"` because for the link itself there's an html `<button>` which is bound to the onclick handler

client/src/components/LinkPicker/LinkPickerTitle.js
emteknetnz: Updated to pass in all attributes

GuySartorelli:
> > While that has been resolved, we now have the situation where there's nothing indicating for screen readers that pressing enter/space on the link element itself will open the edit modal. That's what the aria-role="button" was there for.
> 
> We don't to define `role="button"` because for the link itself there's an html `<button>` which is bound to the onclick handler

Ahhh yup my bad I missed that button somehow 😅

GuySartorelli:
LGTM but has a merge conflict

emteknetnz:
Done

GuySartorelli:
Last lot of conflicts.

GuySartorelli (approved):
LGTM, still works well in combination with the other merged PRs.

Issue #246, opened 2024-03-05
Add deprecation warnings to v3


GuySartorelli:
This was originally part of https://github.com/silverstripe/silverstripe-linkfield/issues/228

As a developer maintaining a project using LinkField v2/v3, I have a a clear path and clear guidance on how to migrate to v4.

## Acceptance Criteria
- [ ] Deprecations warnings are added to v3 for any [public API](https://docs.silverstripe.org/en/5/project_governance/public_api/) which was removed in v4
- [ ] The upgrade guide is updated to say to upgrade to (either `3.x-dev`, `3.0.0`, or `3.0.0-beta2` depending on the below) and resolve any deprecation warnings before migrating.

## To be decided by @maxime-rainville:
- Do we want to tag a `3.0.0-beta2` with the deprecations? Or just leave them on the `3` branch? Or tag stable?

## PR
- https://github.com/silverstripe/silverstripe-linkfield/pull/269
- https://github.com/silverstripe/silverstripe-linkfield/pull/270

maxime-rainville:
I think we need to tag something. My preference would be to tag 3.0.0 stable.

If that's not practical, 3.0.0-beta2 will work.

GuySartorelli:
PRs merged, reassigned to Steve for whatever's left

emteknetnz:
Assigning to @maxime-rainville to decide what to do regarding tagging

maxime-rainville:
Unless someone is telling me that tagging 3.0.0 stable is not advisable, that is still my preferred option.

emteknetnz:
3.0.0 stable was released https://github.com/silverstripe/silverstripe-linkfield/releases/tag/3.0.0

PR #270, opened 2024-04-03
DOC Enable deprecations for linkfield v3

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-linkfield/issues/246

### DIFF
--- a/docs/en/09_migrating/00_upgrading.md
+++ b/docs/en/09_migrating/00_upgrading.md
@@ -6,7 +6,7 @@ summary: A guide for migrating from older versions of silverstripe/linkfield
 # Upgrading from older versions
 
 > [!NOTE]
-> If your site is running Silverstripe CMS 4.x, update your constraint for `silverstripe/linkfield` to `^3.0.0-beta1` and upgrade to CMS 5 first.
+> If your site is running Silverstripe CMS 4.x, update your constraint for `silverstripe/linkfield` to the latest available version of linkfield v3 and upgrade to CMS 5 first.
 > There should be no additional steps required for upgrading from linkfield 2.x to linkfield 3.x.
 > Once you have finished upgrading to CMS 5, return to this guide and continue the linkfield upgrade.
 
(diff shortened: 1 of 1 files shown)


docs/en/09_migrating/00_upgrading.md
```diff
@@ -6,8 +6,9 @@ summary: A guide for migrating from older versions of silverstripe/linkfield
 # Upgrading from older versions
 
 > [!NOTE]
-> If your site is running Silverstripe CMS 4.x, update your constraint for `silverstripe/linkfield` to `^3.0.0-beta1` and upgrade to CMS 5 first.
+> If your site is running Silverstripe CMS 4.x, update your constraint for `silverstripe/linkfield` to the latest available version of linkfield v3 and upgrade to CMS 5 first.
```
ChloeSartorelli, line 9: This is okay if we're tagging it. If we're not tagging it, then "the latest version available" will be interpreted as "the latest tagged version" by most people and they won't use `3.x-dev`.
So just to confirm - are we tagging this thing?

docs/en/09_migrating/00_upgrading.md
```diff
@@ -6,8 +6,9 @@ summary: A guide for migrating from older versions of silverstripe/linkfield
...
 > There should be no additional steps required for upgrading from linkfield 2.x to linkfield 3.x.
+> Enable [deprecation warnings](https://docs.silverstripe.org/en/upgrading/deprecations/) and resolve any deprecation warnings before continuing with this guide.
```
ChloeSartorelli: This note is about upgrading from CMS 4 - did you mean this comment about deprecation warnings to be for people upgrading from linkfield 3.x? If so it needs to be separate.

docs/en/09_migrating/00_upgrading.md
emteknetnz: Sounds like we are? https://github.com/silverstripe/silverstripe-linkfield/issues/246#issuecomment-1979633393 3.0.0-beta2 is absolutely fine since we already have 3.0.0-beta1. I don't have an issue if Max wants to 3.0.0 instead. We can assume we'll tag something.

ChloeSartorelli: Sweet just making sure since our discussion in person yesterday sounded like maybe we wouldn't. If we assume we will then no problem.

docs/en/09_migrating/00_upgrading.md
emteknetnz: Updated

GuySartorelli (approved):
LGTM

Issue #237, opened 2024-02-26
Update linkfield developer doc regarding fluent


maxime-rainville:
## Context 
We've reviewed the high level compatibility of Fluent with LinkField.

We've identified some upstream problems that make it difficult to fully accommodate Fluent with LinkField. Althought the simpler use case seemed to work pretty well.

## User story
As a Fluent developer, I want clear guidance on how to get simple LinkField usage to work with Fluent so I can build interfaces that meet needs of my content authors.

## Acceptance criteria
- The LinkField doc has a section about localisation with Fluent. The section includes.
  - A basic example of how to translate a single Link relation for a page.
- The limitation around translating `has_many` relations is called out
- Links to relevant Fluent doc are added where relevant.

## Related card
- https://github.com/silverstripe/silverstripe-linkfield/issues/210

## PRs
- https://github.com/silverstripe/silverstripe-linkfield/pull/273

GuySartorelli:
The docs for this should really just link to fluent docs - if even that. Really i don't think we need anything more than "`has_one` relations are compatible with fluent, but `has_many` relations are not" and a link to fluent.

maxime-rainville:
I could live with a simple "Fluent compatibility" section that make clear what works and doesn't work with Fluent.

maxime-rainville:
@emteknetnz identified a potential Gridield issue during refinement that would block this card. He will create a matching card for it.

emteknetnz:
These AC's just don't seem good.  This really does not belong in the 5.2.0 change log IMO.

I have this currently ([!Note] section is new):

```md

...

Once the above have all been resolved, you can use that documentation and copy the relevant `BuildTask` code into your projects if you need it. These will be included in `silverstripe/linkfield` directly in the next minor release.

> [!NOTE]
> When using the optional [fluent module](https://github.com/tractorcow-farm/silverstripe-fluent) you can only translate `has_one` single `Link` relations and not the `has_many` multi-relations.
> If using content blocks you can make a `has_many` relation translatable without the `Link` relation itself being translatable. (Each localised instance of the block will have a different `Link` DataObject) 
> Read more about [configuring fluent](https://github.com/tractorcow-farm/silverstripe-fluent/blob/7/docs/en/configuration.md).

Check out [the linkfield documentation](/optional_features/linkfield/) for more information about this module.

...
```

[...]

Issue #210, opened 2024-02-08
SPIKE Review Fluent support for LinkFIeld


maxime-rainville:
We're not sure the LinkField's new implementation will work well with Fluent. The feedback we got from our bespoke devs that use LinkField v3 with Fluent was that in most cases the Link was attached to some parent object that was being translated. In practice, this meant that different language blocks would have different links for each language. 

e.g.: Your create an elemental block with a link. The English block has a link pointing to an English website and the French Block has completely different link pointing to a French website.

Yet the scenario where the Link itself was the thing being translated was judged as desirable and would be considered a substantial shortcoming if it wasn't possible.

e.g. You create a Link pointing to a page. The English and French Link are joint together. User can choose to provide a different title for each Language.

## Timebox
1 day

## TODO
- Validate that the "separate Link in translated block" approach still works as expected.
- Validate if the LinkForm can be made to work with Fluent and allow the creating of translatable Link.
  - Is there a mechanims to copy Link content from one Local to an other.
  - Where a Link is pointing to specific page that has been translated, validate if the Link renders the correct title for the user chosen locale and if following the link URL uses the right locale.
  - If this approach is not working, document possible road block to its implementation.

## Outcome
At the end of this SPIKE, we should:
- know if Link can be translated with Fluent as-is,
- have a rough idea of how much work would be needed to make Link translatable if they are not already,
- be in a position to draft Fluent guidance for LinkField if Link can be translated.

emteknetnz:
## Summary:

Fluent seems to work fine for has_one LinkField's, however it is rough for has_many MutliLinkField's

## Shortcomings:

Things work fairly well for the has_one LinkField, but seem to fall apart pretty badly for the has_many MutliLinkField. I'm not sure if that's because fluent seems to have not so great support for has_many (and many_many) relations, or if I just don't know how to properly use fluent

The only real shortcoming with the has_one was there's a 'Localisation' tab with a gridfield that's supposed to render, though because we're using a react modal there is no react gridfield that can be rendered.  I need to add this to GridField.php to prevent a fatal exception:

```php
// GridField.php
// ...
protected $schemaDataType = self::SCHEMA_DATA_TYPE_STRUCTURAL;
```

And it simply renders as nothing:

![image](https://github.com/silverstripe/silverstripe-linkfield/assets/4809037/2dd7d6b4-e932-42dd-8575-f107dfab9f9e)

Possibly we could just manually remove the tab since its' non-functional

[...]

PR #273, opened 2024-04-10
DOC Document localising links

GuySartorelli:
## Description

Documents localising links with fluent.

Note that this won't work unless https://github.com/silverstripe/silverstripe-framework/pull/11195 is also merged. You can work around that by adding this configuration:

```yml
SilverStripe\LinkField\Models\Link:
  extensions:
    - TractorCow\Fluent\Extension\FluentVersionedExtension
  field_exclude:
    - OwnerRelation
```

## Original fluent spike
- https://github.com/silverstripe/silverstripe-linkfield/issues/210

## Issues

- https://github.com/silverstripe/silverstripe-linkfield/issues/237

### DIFF
--- a/docs/en/02_configuration.md
+++ b/docs/en/02_configuration.md
@@ -129,3 +129,27 @@ class MyCustomLink extends Link
     // ...
 }
 ```
+
+## Localisation with `tractorcow/silverstripe-fluent` {#localisation}
+
+If you install the optional `tractorcow/silverstripe-fluent` module, you can localise your links by adding `FluentVersionedExtension` to the `Link` model. This works for all links, regardless of whether they're in a `has_one` or `has_many` relation.
+
+```yml
+SilverStripe\LinkField\Models\Link:
+  extensions:
+    - TractorCow\Fluent\Extension\FluentVersionedExtension
+```
+
+If you have chosen to remove the [`Versioned`](api:SilverStripe\Versioned\Versioned) extension from the `Link` model, apply `FluentExtension` instead:
+
+```yml
+SilverStripe\LinkField\Models\Link:
+  extensions:
+    - TractorCow\Fluent\Extension\FluentExtension
+```
+
+> [!NOTE]
+> Because `GridField` doesn't work in a react context, you won't have access to some UI functionality such as the "Localisation" tab.
+> The "Localisation" tab is referred to in fluent documentation as the "localisation manager".
+
+See the documentation in the [`tractorcow-farm/silverstripe-fluent` GitHub repository](https://github.com/tractorcow-farm/silverstripe-fluent) if you need to do anything more complex.


emteknetnz (changes requested):
I had a quick test of this after referencing my notes in the [original spike](https://github.com/silverstripe/silverstripe-linkfield/issues/210#issuecomment-1943029385)

While has_many looks like it works because, it really doesn't
- Sorting has_many in one locale will unexpectedly sort the links in the other locale which is very unexpected
- Archiving a link does nothing. Try refreshing the page after deleting it.

I stopped testing at this point. I think it's simply a case that "has_many doesn't work"

GuySartorelli:
> Sorting has_many in one locale will unexpectedly sort the links in the other locale which is very unexpected

I disagree - I think that is very expected. I think in most cases you would want the links in the same order for all locales and only change the text. Imagine having to go through all of your locales every time you want to change the order of links - what a nightmare!

If developers want to sort the links differently per locale, that is doable but goes beyond the basic scenario I have documented here - people who want that probably are already doing that kind of thing elsewhere in their project as well - but if not, they can read the fluent docs to look for information about how fluent works.

I want to avoid duplication of documentation as much as possible.

> Archiving a link does nothing. Try refreshing the page after deleting it.

It archives the _specific version you have in that locale_. After you refresh, you see the version of the link which is stored in the fallback locale. Notice the title of the link is different before and after archiving in this video:

[Screencast from 10-04-24 16:19:02.webm](https://github.com/silverstripe/silverstripe-linkfield/assets/36352093/8a72c574-ae86-422b-86a8-54184842de90)

There are a couple of UX issues here:
1. The link shouldn't disappear after archiving - it should just immediately show the copy of the link from the fallback locale
2. There should be fluent labels on the links indicating *where* the link information is coming from (so it's clear that version is not from **this** locale, it's from the fallback)
    - Obviously we shouldn't add coupling here - but we could provide an API for adding arbitrary labels that fluent or some intermediary "compatibility" module could take advantage of
1. The archive option shouldn't be available when the link hasn't been saved in the current locale

Those UX issues are not in the scope of this PR, which is just adding documentation. If you want, I can document these UX issues.

> I think it's simply a case that "has_many doesn't work"

I disagree. As stated above, it seems like there are just a few UX issues that need to be ironed out.
The archiving UX problem is not specific to `has_many` either - that's the behaviour for all links, even `has_one`.

emteknetnz:
If we want to say that fluent works correctly, though there's some caveats, and we need to do more work on linkfield to get fluent to work as you think it should, then it's too soon to merge these docs as they are where "fluent has_many works" where IMO it really doesn't

I think this one needs to go back in refinement for discussion to work out what we want to do with fluent since it sounds like there's a bunch more work to do.

GuySartorelli:
You're focusing a lot on `has_many` but the only outstanding issues seem to be around archiving, which as I noted affect `has_one` as well.
In any case, I've put the issue back into refinement and I'll close this PR.

Issue #276, opened 2024-04-18
Child LinkField without a link will show an error after inline save


emteknetnz:
After inline saving an element, a child LinkField without any links will make an XHR request that will 404 and front end will show "Failed to load link(s)"

## Replication steps
- Have an elemental block with a "has on" to link and matching link field
- Create the block.
- Edit the block and save without creating a link

Results:  XHR request that will 404 and front end will show "Failed to load link(s)"

Wasn't tested with has many.

## Acceptance criteria
- Fix the bug
- Add a behat test to make sure elemental doesn't crash if you don't save a link in your linkfield.

## PRs
- https://github.com/silverstripe/silverstripe-linkfield/pull/278

GuySartorelli:
@emteknetnz Please also add a behat test - the jest test is good for testing the specific value is handled, but it doesn't validate that the scenario described actually works. If something changes about how elemental handles forms or fields or values or form submission or any other number of things, this could break in another way.

emteknetnz:
Both linkfield and elemental are optional modules with no dependency between them, I'm not sure how you'd even create a behat test here. It would probably need to be on kitchen-sink which seems pretty wrong. I think it's fine not having a behat test here.

GuySartorelli:
Linkfield has elemental as a dev dependency, and already has a [feature for behat tests with elemental](https://github.com/silverstripe/silverstripe-linkfield/blob/4/tests/behat/features/linkfield-in-elemental.feature). So it would fit in nicely there.

If with that context you still think it's not worth doing then feel free to close the issue.

emteknetnz:
Will close

PR #278, opened 2024-05-02
FIX Handle a string "0" value

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-linkfield/issues/276

Note that that AC says create a behat test, though I've used a jest test instead because it's more appropriate. The jest test will fail without the code fix

I've confirmed that this issue only affected single linkfield and not multilink field


GuySartorelli (approved):
LGTM

PR #309, opened 2024-07-30
DEP Remove unused apollo dependency

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/1799

GraphQL is unused in Linkfield V4, so there is no need for apollo

Note: unlike other PRs on the parent issue which target CMS 6, this one targets CMS 5.2


GuySartorelli (approved):
LGTM

Issue #291, opened 2024-05-27
`canArchive()` is not called before archiving versioned links


GuySartorelli:
Note that this has been looked at from a security standpoint, and the CMS Squad has decided it does not constitute a security issue under our current security process.

`‎‎LinkFieldController::linkDelete()` checks `canDelete()`, but for versioned records it doesn't check `canArchive`. It does have version-aware logic to call `doArchive()` instead of `delete()`, so it should also use the correct permission check.

## Acceptance criteria
- For versioned links, the `canArchive()` permission method is called
- For non-versioned links, the `canDelete()` permission method is called

## PRs
- https://github.com/silverstripe/silverstripe-linkfield/pull/315

GuySartorelli:
PR merged

PR #315, opened 2024-08-06
FIX Check canArchive() permission instead of canDelete()

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-linkfield/issues/291

### DIFF
--- a/tests/php/Controllers/LinkFieldControllerTest.php
+++ b/tests/php/Controllers/LinkFieldControllerTest.php
@@ -11,6 +11,7 @@ use SilverStripe\Control\Session;
 use SilverStripe\LinkField\Controllers\LinkFieldController;
 use SilverStripe\LinkField\Tests\Models\LinkTest\LinkOwner;
 use SilverStripe\Versioned\Versioned;
+use SilverStripe\LinkField\Models\Link;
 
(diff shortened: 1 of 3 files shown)


tests/php/Controllers/LinkFieldControllerTest.php
```diff
@@ -570,9 +570,12 @@ public function provideLinkDelete(): array
                 'fail' => '',
                 'expectedCode' => 204,
             ],
-            'Reject fail canDelete()' => [
+            // note there isn't a canDelete() test here because it seems impossible to get an
+            // unversioned Link because there's no way to actually remove the Versioned extension
+            // from any subclass of Link since we're unable to the use app/_config.php method
```
ChloeSartorelli: Why can't you just use `MyLinkSubclass::remove_extension(Versioned::class);` in the test?

GuySartorelli (changes requested):

tests/php/Controllers/LinkFieldControllerTest.php
emteknetnz: Updated

GuySartorelli (approved):
LGTM

PR #316, opened 2024-08-12
MNT Add extension in finally block

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-linkfield/issues/317

Add the Versioned extension back in using a finally block to prevent any bleed over to other tests in case the test fails

Also upgrades `==` to standard `===`


GuySartorelli (approved):
Makes sense. LGTM

PR #319, opened 2024-08-13
ENH Don't emit deprecation warnings for unavoidable API calls

GuySartorelli:
`Versioned::canArchive()` is deprecated in https://github.com/silverstripe/silverstripe-versioned/pull/461

We can't stop calling it directly until CMS 6 because people may have implemented extensions that update the result of this permission check.

## Issue
- https://github.com/silverstripe/silverstripe-versioned/issues/447

### DIFF
--- a/src/Controllers/LinkFieldController.php
+++ b/src/Controllers/LinkFieldController.php
@@ -16,6 +16,7 @@ use SilverStripe\ORM\ValidationResult;
 use SilverStripe\Control\Controller;
 use SilverStripe\Control\HTTPRequest;
 use SilverStripe\Core\Injector\Injector;
+use SilverStripe\Dev\Deprecation;
 use SilverStripe\Forms\HiddenField;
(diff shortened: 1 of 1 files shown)


src/Controllers/LinkFieldController.php
```diff
@@ -133,7 +134,7 @@ public function linkDelete(): HTTPResponse
         }
         $link = $this->linkFromRequest();
         if ($link->hasExtension(Versioned::class)) {
-            if (!$link->canArchive()) {
+            if (!Deprecation::withNoReplacement(fn() => $link->canArchive())) {
```
```suggestion
            $canArchive = Deprecation::withNoReplacement(fn() => $link->canArchive());
            if (!$canArchive) {
```

emteknetnz (changes requested):

src/Controllers/LinkFieldController.php
ChloeSartorelli: Done

emteknetnz (approved):

PR #320, opened 2024-08-19
DEP Limit PHP support for CMS 6

GuySartorelli:
## Issue
- https://github.com/silverstripe/.github/issues/297

### DIFF
--- a/tests/php/Tasks/LinkFieldMigrationTaskTest.php
+++ b/tests/php/Tasks/LinkFieldMigrationTaskTest.php
@@ -3,8 +3,8 @@
 namespace SilverStripe\LinkField\Tests\Tasks;
 
 use LogicException;
+use ReflectionClass;
 use ReflectionMethod;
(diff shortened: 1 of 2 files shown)


tests/php/Tasks/LinkFieldMigrationTaskTest.php
ChloeSartorelli, line 1: Changes to this file are needed to fix https://github.com/silverstripe/silverstripe-linkfield/actions/runs/10447675451/job/28980857139?pr=320
> Calling ReflectionProperty::setValue() with a single argument is deprecated

https://www.php.net/manual/en/reflectionproperty.setvalue.php says
> Note: As of PHP 8.3.0, calling this method with a single argument is deprecated, use [ReflectionClass::setStaticPropertyValue()](https://www.php.net/manual/en/reflectionclass.setstaticpropertyvalue.php) instead.

emteknetnz (approved):

PR #326, opened 2024-09-04
FIX Use Updated names of top page extensions

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-elemental/issues/1240

### DIFF
--- a/src/Extensions/UsedOnTableExtension.php
+++ b/src/Extensions/UsedOnTableExtension.php
@@ -3,7 +3,8 @@
 namespace SilverStripe\LinkField\Extensions;
 
 use DNADesign\Elemental\Models\BaseElement;
-use DNADesign\Elemental\TopPage\DataExtension as TopPageExtension;
+use DNADesign\Elemental\Extensions\TopPageElementExtension;
+use DNADesign\Elemental\Extensions\TopPageFluentElementExtension;
 use SilverStripe\Core\Extension;
 use SilverStripe\LinkField\Models\FileLink;
 use SilverStripe\ORM\DataObject;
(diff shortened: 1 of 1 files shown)


src/Extensions/UsedOnTableExtension.php
```diff
@@ -23,7 +24,11 @@ protected function updateUsageAncestorDataObjects(array &$ancestorDataObjects, D
         if (!class_exists(BaseElement::class) || !is_a($owner, BaseElement::class)) {
             return;
         }
-        $page = $owner->hasExtension(TopPageExtension::class) ? $owner->getTopPage() : $owner->getPage();
+        if ($owner->hasExtension(TopPageElementExtension::class) || $owner->hasExtension(TopPageFluentElementExtension::class)) {
+            $page = $owner->getTopPage();
+        } else {
+            $page = $owner->getPage();
+        }
```
ChloeSartorelli, line 31: I prefer the ternary... but it's not worth ping pong.

ChloeSartorelli: But the change to logic isn't needed as discussed irl. Please remove the unnecessary addition

GuySartorelli (changes requested):

emteknetnz:
Have validated locally that hasExtension() will perform an is_a() check that will check for subclasses. Don't need this PR

emteknetnz:
Actually we do :laughing: renamed the class, re-opening, and have removed the check for the fluent class

emteknetnz:
Wouldn't let me reopen as I had force pushed over the last commit, and opened this instead https://github.com/silverstripe/silverstripe-linkfield/pull/327

PR #329, opened 2024-09-06
API Update API to reflect changes to CLI interaction

GuySartorelli:
Reflects changes in https://github.com/silverstripe/silverstripe-framework/pull/11353

## Issue
- https://github.com/silverstripe/silverstripe-framework/issues/11341


src/Tasks/GorriecoeMigrationTask.php
```diff
@@ -13,18 +13,18 @@
 use SilverStripe\ORM\DB;
 
 /**
- * @deprecated 4.0.0 Will be removed without equivalent functionality.
+ * @deprecated 5.0.0 Will be removed without equivalent functionality.
```
emteknetnz, line 16: Shouldn't this just be removed like the linkable migration task?

emteknetnz (changes requested):

src/Tasks/GorriecoeMigrationTask.php
ChloeSartorelli: No, because the module is fairly likely to be made compatible with CMS 6 so people may still want to migrate off it in CMS 6. IMO we shouldn't remove it until we get to a point where it's obvious it won't be available as something people can use and migrate off from.

emteknetnz (approved):

PR #331, opened 2024-09-09
MNT Resolve deprecation warning in test

GuySartorelli:
See https://www.php.net/manual/en/migration83.deprecated.php about the deprecation
See https://www.php.net/manual/en/reflectionproperty.setvalue.php for the suggested solution

Not targeting `4.0` because this code doesn't exist on that branch.

## Issues
- https://github.com/silverstripe/.github/issues/309


GuySartorelli:
Replacing with https://github.com/silverstripe/silverstripe-linkfield/pull/332

PR #332, opened 2024-09-10
MNT Resolve deprecation warning in test

GuySartorelli:
See https://www.php.net/manual/en/migration83.deprecated.php about the deprecation
See https://www.php.net/manual/en/reflectionproperty.setvalue.php for the suggested solution

Not targeting `4.0` because this code doesn't exist on that branch.

## Issues
- https://github.com/silverstripe/.github/issues/309

### DIFF
--- a/tests/php/Tasks/LinkFieldMigrationTaskTest.php
+++ b/tests/php/Tasks/LinkFieldMigrationTaskTest.php
@@ -3,6 +3,7 @@
 namespace SilverStripe\LinkField\Tests\Tasks;
 
 use LogicException;
+use ReflectionClass;
 use ReflectionMethod;
 use ReflectionProperty;
 use RuntimeException;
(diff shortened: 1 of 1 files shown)


GuySartorelli:
Prefer lowest failure is unrelated

emteknetnz (changes requested):
We're missing the `setAccessible(true)` which should still be required according to [this comment](https://www.php.net/manual/en/reflectionclass.setstaticpropertyvalue.php#114740)

isCliOverride is a private static

Is the linked comment wrong? Or is the unit test just wrong?

GuySartorelli:
The comment says 
> Calling this method on a static property that is not public will return a ReflectionException stating the property does not exist

That's not happening, so it must be false.
It's from 10 years ago, things have likely changed since then

emteknetnz (approved):

Issue #333, opened 2024-09-10
CMS 5.3.0-beta1 regression: "Failed to load link" with anchor links


GuySartorelli:
When adding a link to a "page on the site" with `silverstripe/linkfield` inside an elemental block, we get this error when publishing (unclear if it's publishing the page or the block)

> Failed to load link(s)

![image](https://github.com/user-attachments/assets/121d5d26-6698-49ab-b95b-ffcd2156c1c7)

Note that the correct link displays after refreshing the page.

## Steps to reproduce

Note this assumes you have `silverstripe/frameworktest` installed, along with the rest of kitchen sink.

1. Add Link List Block
3. Go to Single Link 
5. Select Add Link
7. Select Page on the site from dropdown
9. Select a page
11. Add Anchor
13. Select Update link
15. Link should display in DRAFT mode
17. Select Publish

## PRs
- https://github.com/silverstripe/silverstripe-linkfield/pull/335

GuySartorelli:
Can reproduce this - though some of the steps are a bit odd (why does it say "update link" in the instructions when you haven't even created it, for example?)

GuySartorelli:
Frustratingly cannot reproduce this locally.
The environment I am able to reproduce it in is https://cmstestm-prod.sites.silverstripe.com/ which is PHP 8.3

GuySartorelli:
I suspect it is a result of this:

![image](https://github.com/user-attachments/assets/483d1829-53fe-4e37-b1d4-794edf75f4eb)

The correct URL is https://cmstestm-prod.sites.silverstripe.com/admin/linkfield/data?itemIDs[]=95 (which it does eventually try) but for some reason the ID is split into two for the first request (https://cmstestm-prod.sites.silverstripe.com/admin/linkfield/data?itemIDs[]=9&itemIDs[]=5)

I'll try reproducing locally with IDs in the double digits.

GuySartorelli:
Can reproduce locally now - getting the exact same split ID problem when running in my local with double-digit IDs.

emteknetnz:
Linked PR merged

Issue #256, opened 2024-03-14
CMS 5.2.0-beta1. Regression test: LinkField has "Draft / Modified" label even when Elemental Block was published


sabina-talipova:
## Description
Executed on: SC PHP 8.3 Chrome LinkField
Test scenario: https://studio.cucumber.io/projects/301855/test-runs/947336/folder-snapshots/11144442/scenario-snapshots/34759621/test-snapshots/46444047

After publishing block draft label won't go away, but it has been published on the front end. when the page refreshes only it will go away.

### Steps to reproduce issue:
- I create a new Elemental block with LinkField;
- I create a new Link;
- I see "Draft" badge in LinkField;
- I publish the block (not a Page);
-  I still see "Draft" badge in LinkField;
-  I refresh the Page and then a badge disappears.

## Acceptance criteria
- Manually publishing an elemental block with a owned Link relation update the status of the relevant linkfield.

## PR
- https://github.com/silverstripe/silverstripe-linkfield/pull/268
- https://github.com/silverstripe/silverstripe-elemental/pull/1161
- https://github.com/silverstripe/silverstripe-admin/pull/1771
- https://github.com/silverstripe/silverstripe-elemental/pull/1200
- https://github.com/silverstripe/silverstripe-linkfield/pull/294

sabina-talipova:
There is the similar problem for another nested versioned DataObject.
For example for files:
<img width="636" alt="Screenshot 2024-03-27 at 12 55 40 PM" src="https://github.com/silverstripe/silverstripe-linkfield/assets/87288324/1fbc1862-f1b3-4209-9c35-9b6b082f6425">

**Update:** Related issue was open https://github.com/silverstripe/silverstripe-asset-admin/issues/1450

emteknetnz:
@sabina-talipova I've closed the linked PRs because I don't think they're going about solving this the correct way.  It's doing a couple of things I dislike:
- Creating a coupling between two optional modules
- Add  client side form state thus creating a new source of truth. This adds complexity and can easily lead to creating regressions later on. It's FAR easier to only use the server as the source of truth.
- It's also not solving the [asset-admin issue](https://github.com/silverstripe/silverstripe-asset-admin/issues/1450) noted above which has the same symptom

Instead what I think we want to do is force the inline form (created using FormBuilder which uses redux-form) to refresh after a successful publish/unpublish so that linkfield will subsequently refresh its own data from the server. 

[...]

PR #335, opened 2024-09-10
FIX Don't split ID into separate digits

GuySartorelli:
The bug is only seen when you have multi-digit IDs. For some reason the ID was coming through as a string, so it didn't get transformed into an array, so we ended up iterating over the string and therefore requesting info for each digit.

## Issue
- https://github.com/silverstripe/silverstripe-linkfield/issues/333


emteknetnz (approved):
Tested locally, things seem to work correctly, at least in terms of fixing this bug which I was able to replicate.

Note that after inline publishing the block doesn't update its versioned badge to show published, though given that's a separate issue we can merge this as is

GuySartorelli:
The versioned badge issue was raised as a regression for 5.2.0: https://github.com/silverstripe/silverstripe-linkfield/issues/256

PR #338, opened 2024-09-24
ENH Use symfony/validation logic

GuySartorelli:
Relies on changes in https://github.com/silverstripe/silverstripe-framework/pull/11399

## Issue
- https://github.com/silverstripe/.github/issues/193

### DIFF
--- a/src/Models/ExternalLink.php
+++ b/src/Models/ExternalLink.php
@@ -28,6 +28,7 @@ class ExternalLink extends Link
     {
         $this->beforeUpdateCMSFields(function (FieldList $fields) {
             $field = UrlField::create('ExternalUrl', _t(__CLASS__ . '.EXTERNAL_URL_FIELD', 'External URL'));
+            $field->setAllowedProtocols(['http', 'https']);
             $field->setDescription(_t(
                 __CLASS__ . '.EXTERNAL_URL_FIELD_DESCRIPTION',
                 'The URL must start with either http:// or https://'


src/Models/ExternalLink.php
```diff
@@ -28,6 +28,7 @@ public function getCMSFields(): FieldList
     {
         $this->beforeUpdateCMSFields(function (FieldList $fields) {
             $field = UrlField::create('ExternalUrl', _t(__CLASS__ . '.EXTERNAL_URL_FIELD', 'External URL'));
+            $field->setAllowedProtocols(['http', 'https']);
```
ChloeSartorelli, line 31: Explicitly enforcing these protocols, even if the global config allows more protocols than this. We had decided originally this was explicitly what we wanted this link type to be used for. We can always loosen it later on if we want to.

emteknetnz (approved):

emteknetnz (changes requested):
Behat failure

GuySartorelli:
Oops, I somehow missed a reference to the old `PasswordValidator`.
Should be fixed by https://github.com/silverstripe/silverstripe-behat-extension/pull/286

emteknetnz (changes requested):
Still failing after merging behat-extension PR and re-running

GuySartorelli:
I must have been very distracted when doing that original framework PR lol. This behat failure is caused by a bug I introduced there.
Fixed by https://github.com/silverstripe/silverstripe-framework/pull/11414

emteknetnz (approved):

PR #342, opened 2024-10-14
ENH Use config instead of runtime code to remove menu item

GuySartorelli:
Quick tidyup to use the `ignore_menuitem` configuration property instead of removing the menu item at runtime.

Can't refactor to use `AdminController` instead of `LeftAndMain` yet because it needs formschema - but this small refactor will make that slightly easier when I come to it.

## Issue
- https://github.com/silverstripe/silverstripe-admin/issues/1761

### DIFF
--- a/_config.php
+++ b/_config.php
@@ -1,6 +1,0 @@
-<?php
-
-use SilverStripe\Admin\CMSMenu;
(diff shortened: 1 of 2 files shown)


GuySartorelli:
CI failure is unrelated

michalkleiner (approved):

Issue #346, opened 2024-11-11
Missing elemental method on AbstractLinkField


emteknetnz:
After merging https://github.com/silverstripe/silverstripe-elemental/issues/1268 there's now an issue as the old method no longer exists

```
1) SilverStripe\LinkField\Tests\Form\AbstractLinkFieldTest::testElementalNamespaceRemoved
Error: Call to undefined method DNADesign\Elemental\Controllers\ElementalAreaController::removeNamespacesFromFields()
```

## PRs
- https://github.com/silverstripe/silverstripe-linkfield/pull/347

GuySartorelli:
PR merged

PR #347, opened 2024-11-11
FIX Call updated method

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-linkfield/issues/346

### DIFF
--- a/src/Form/AbstractLinkField.php
+++ b/src/Form/AbstractLinkField.php
@@ -3,13 +3,16 @@
 namespace SilverStripe\LinkField\Form;
 
 use DNADesign\Elemental\Controllers\ElementalAreaController;
+use DNADesign\Elemental\Forms\EditFormFactory;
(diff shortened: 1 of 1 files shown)


src/Form/AbstractLinkField.php
```diff
@@ -170,10 +173,11 @@ protected function getOwnerFields(): array
         if (class_exists(BaseElement::class) && is_a($owner, BaseElement::class)) {
             // Remove namespaces from inline editable blocks
             // This will return an empty array for non-inline editable blocks (e.g. blocks in a gridfield)
```
ChloeSartorelli: This comment doesn't seem to be correct anymore

GuySartorelli (changes requested):

src/Form/AbstractLinkField.php
emteknetnz: Updated

GuySartorelli (approved):
LGTM

PR #348, opened 2024-11-12
API Make LinkFieldController a subclass of FormSchemaController

GuySartorelli:
Needs https://github.com/silverstripe/silverstripe-admin/pull/1850 for CI to go green. Please merge that first.

## Issue
- https://github.com/silverstripe/silverstripe-admin/issues/1762

### DIFF
--- a/tests/php/Controllers/LinkFieldControllerTest.php
+++ b/tests/php/Controllers/LinkFieldControllerTest.php
@@ -251,9 +251,9 @@ class LinkFieldControllerTest extends FunctionalTest
         $response = $this->post($url, $data, $headers);
         $this->assertSame($expectedCode, $response->getStatusCode());
         if ($fail === 'csrf-token') {
-            // Will end up at an HTML page with "Silverstripe - Bad Request"
-            $this->assertSame('text/html; charset=utf-8', $response->getHeader('Content-type'));
-            $this->assertStringContainsString('Silverstripe - Bad Request', $response->getBody());
+            // Gives suitable error message for XHR request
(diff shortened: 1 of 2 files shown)


tests/php/Controllers/LinkFieldControllerTest.php
```diff
@@ -251,9 +251,9 @@ public function testLinkFormPost(
         $response = $this->post($url, $data, $headers);
         $this->assertSame($expectedCode, $response->getStatusCode());
         if ($fail === 'csrf-token') {
-            // Will end up at an HTML page with "Silverstripe - Bad Request"
-            $this->assertSame('text/html; charset=utf-8', $response->getHeader('Content-type'));
-            $this->assertStringContainsString('Silverstripe - Bad Request', $response->getBody());
+            // Gives suitable error message for XHR request
+            $this->assertStringStartsWith('text/plain', $response->getHeader('Content-type'));
+            $this->assertStringContainsString('There seems to have been a technical problem', $response->getBody());
```
ChloeSartorelli, line 256: The modal links used to be routed through `LeftAndMain` which meant the error handling in that class was used. For some reason they weren't picked up as XHR requests so the full HTML error page was being rendered which wasn't really the correct behaviour, but it's what happened.

Now, a simple `400` response (the default one produced when a form's CSRF token is missing) is produced and isn't altered. This is the correct behaviour.

The end result for the user is no change.

emteknetnz (approved):

emteknetnz (approved):

PR #351, opened 2024-11-20
ENH Use getStatusFlags() instead of hardcoded statuses

GuySartorelli:
Status flags are made generic in https://github.com/silverstripe/silverstripe-framework/pull/11460

This PR uses the new `getStatusFlags()` method instead of building hardcoded status flags.

Ends up looking something like this (custom flag added by overriding `getStatusFlags()`)
Screenshot from 2024-11-20 14-56-56

## Issue
- https://github.com/silverstripe/silverstripe-cms/issues/2947


client/src/components/LinkPicker/LinkPicker.scss
```diff
@@ -198,9 +198,11 @@
   width: 100%;
 
   .badge {
-    color: #cf3f00;
-    background-color: #fff2ea;
```
ChloeSartorelli, line 202: Rely on styling in admin for colours

client/src/components/LinkPicker/LinkPicker.scss
```diff
@@ -198,9 +198,11 @@
...
     padding: 2px 3px 2px 4px;
+
+    &:not(:last-child) {
+      margin-right: 5px;
+    }
```
ChloeSartorelli, line 205: Needs a gap between badges - this value matches the right margin on the title text

src/Models/Link.php
```diff
@@ -253,7 +254,7 @@ public function forTemplate(): string
     {
         // First look for a subclass of the email template e.g. EmailLink.ss which may be defined
         // in a project. Fallback to using the generic Link.ss template which this module provides
-        return $this->renderWith([static::class, Link::class]);
+        return (string) $this->renderWith([static::class, Link::class]);
```
ChloeSartorelli, line 257: Unrelated change - my IDE was upset that the return type was string but this returns `DBHtmlText` - probably just because we're using strict mode here when we don't use it anywhere else.

client/src/components/LinkPicker/LinkPickerTitle.js
```diff
@@ -17,27 +17,30 @@ const stopPropagation = (fn) => (e) => {
...
+  }
+  return badges;
 };
 
 const LinkPickerTitle = ({
   id,
   title,
   description,
   versionState,
+  statusFlags,
```
emteknetnz, line 43: How is this prop being passed in?  There's no corresponding PHP change in this PR?

emteknetnz (changes requested):

client/src/components/LinkPicker/LinkPickerTitle.js
ChloeSartorelli: Yes there is. See the `Link.php` file.
https://github.com/silverstripe/silverstripe-linkfield/pull/351/files#diff-10018e9f26d86c4bbc71cb5f4e152ce6566f03cf1f40e9ec606f125371a60ad5R241

emteknetnz (approved):

Issue #359, opened 2025-01-05
Plural and Singular CMS Message update


n8-dev:
### Description

`/client/lang/en.js`

```diff
- "LinkField.SAVE_RECORD_FIRST": "Cannot add links until the record has been saved",
+ "LinkField.SAVE_RECORD_FIRST": "Cannot add link(s) until the record has been saved",
```

Client was confused about the usage on `links` when Linkfield was used via a simple has_one

### Validations

- [X] You intend to implement the feature yourself
- [X] You have read the [contributing guide](https://docs.silverstripe.org/en/contributing/code/)
- [X] You strongly believe this feature should be in core, rather than being its own community module
- [X] You have checked for existing issues or pull requests related to this feature (and didn't find any)

GuySartorelli:
PR merged. This will be included in the April minor release.

PR #360, opened 2025-01-06
Change Links to Link(s) to match the rest of the messages

n8-dev:
## Description
Tweaked messaging to handle the situation when context is both Plural and Singular.
Much like how it already is here.
```js
    "LinkField.FAILED_TO_LOAD_LINKS": "Failed to load link(s)",
```

i.e.
```diff
- "LinkField.SAVE_RECORD_FIRST": "Cannot add links until the record has been saved",
+ "LinkField.SAVE_RECORD_FIRST": "Cannot add link(s) until the record has been saved",
```

## Manual testing steps

Update text should show when loading an item that has a relation to 1/many linkfields that isn't saved and you wish to add a link to it.

## Issues

- #359

## Pull request checklist

- [x] All commits are relevant to the purpose of the PR (e.g. no debug statements, unrelated refactoring, or arbitrary linting)
- [x] The commit messages follow our [commit message guidelines](https://docs.silverstripe.org/en/contributing/code/#commit-messages)
- [x] The PR follows our [contribution guidelines](https://docs.silverstripe.org/en/contributing/code/)
- [x] Code changes follow our [coding conventions](https://docs.silverstripe.org/en/contributing/coding_conventions/)
- [x] This change is covered with tests (or tests aren't necessary for this change)
- [x] Any relevant User Help/Developer documentation is updated; for impactful changes, information is added to the changelog for the intended release
- [x] CI is green


n8-dev:
Working on updating the dist file

GuySartorelli (changes requested):
Please remove all of the changes to localised files for languages other than English - those are managed in transifex. Feel free to [update the references in transifex](https://docs.silverstripe.org/en/5/contributing/translations/) if you like, they'll get pulled down whenever our next localisation run is (I think those are every 3 months if I'm remembering correctly)

GuySartorelli:
Actually, looking at this - we should probably have two separate messages. The javascript component knows whether this is a multi-link field or a single link field, so we should always use the existing "links" message for a multi link field and use a new "link" message for single link fields.

client/lang/en.js
```diff
@@ -23,7 +23,8 @@ if (typeof(ss) === 'undefined' || typeof(ss.i18n) === 'undefined') {
     "LinkField.LINK_DRAFT_TITLE": "Link has draft changes",
     "LinkField.LINK_MODIFIED_LABEL": "Modified",
     "LinkField.LINK_MODIFIED_TITLE": "Link has unpublished changes",
-    "LinkField.SAVE_RECORD_FIRST": "Cannot add links until the record has been saved",
+    "LinkField.SAVE_RECORD_FIRST": "Cannot add link until the record has been saved",
+    "LinkField.SAVE_RECORD_FIRST_MULTI": "Cannot add links until the record has been saved",
```
ChloeSartorelli: We should avoid changing existing localisation keys because we don't use version control in transifex, so this could change the localisation key from _previous_ versions of linkfield as well if their localisations are updated.

It might also trip anyone up who's using this key for their own purposes.

Instead, just keep using the old key for the text that isn't changing and only add a new one.
```suggestion
    "LinkField.SAVE_RECORD_FIRST": "Cannot add links until the record has been saved",
    "LinkField.SAVE_RECORD_FIRST_SINGLE": "Cannot add link until the record has been saved",
```
You'll need to update the JSON and LinkField.js as well.

client/src/components/LinkField/LinkField.js
```diff
@@ -524,7 +524,9 @@ const LinkField = ({
   const renderPicker = !loadingError && !inHistoryViewer && !saveRecordFirst && (isMulti || linkIDs.length === 0);
   const renderModal = !loadingError && !saveRecordFirst && Boolean(editingID);
   const loadingErrorText = i18n._t('LinkField.FAILED_TO_LOAD_LINKS', 'Failed to load link(s)');
-  const saveRecordFirstText = i18n._t('LinkField.SAVE_RECORD_FIRST', 'Cannot add links until the record has been saved');
+  const saveRecordFirstText = isMulti 
+    ? i18n._t('LinkField.SAVE_RECORD_FIRST_MULTI', 'Cannot add links until the record has been saved')
+    : i18n._t('LinkField.SAVE_RECORD_FIRST', 'Cannot add link until the record has been saved');
```
```suggestion
    ? i18n._t('LinkField.SAVE_RECORD_FIRST', 'Cannot add links until the record has been saved')
    : i18n._t('LinkField.SAVE_RECORD_FIRST_SINGLE', 'Cannot add link until the record has been saved');
```

n8-dev:
Linkfield isn't in https://explore.transifex.com/silverstripe/ ?

GuySartorelli:
Weird. It's definitely in the app: https://app.transifex.com/silverstripe/silverstripe-linkfield/dashboard/
I'll have to look into how that explore page gets its info

n8-dev:
Cool, can see that, is there anything specific to do with the new option added?

GuySartorelli:
Nah, that'll get collected and added to transifex when we do our next localisation round.

GuySartorelli (changes requested):

client/lang/en.js
n8-dev: Ah dang.

Makes sense there.

Too big of change...

I suspect Transifex would have some sort of auto invalidate certain things mechanic but yeah, less work this way.

n8-dev:
Think this is ready.

Unsure about a few things...
- ENH vs TRN
- I think this is the right target based on change level...
- Is CI green?

GuySartorelli:
- `ENH` is correct here since you're enhancing the functionality to include a second separate message, you're not _only_ affecting the localisation strings and keys.
  - Usually the colon shouldn't be included, i.e. it should be `ENH My commit message here.` not `ENH: My commit message here.`. It doesn't matter much really, but since you need to make changes anyway (see below about reverted change) you may as well fix that at the same time.
- Target branch is correct
- CI doesn't look like it has run for some reason (you can see in the "checks" tab at the top, it's got 0 items) - GitHub does this sometimes. Another force push usually gets it to right itself.

You've reverted https://github.com/silverstripe/silverstripe-linkfield/pull/360#discussion_r1903386320 and https://github.com/silverstripe/silverstripe-linkfield/pull/360#discussion_r1903386519 in your commit squashing efforts, though.

n8-dev:
Dang!

Onwards!

n8-dev:
Hmm, workflow awaiting approval?

https://github.com/silverstripe/silverstripe-linkfield/actions/runs/12624941936

This because it's a fork PR and something isn't connected quite right?

n8-dev:
Now its running idk 🤷

GuySartorelli:
Oh yeah, I forgot the new merge experience doesn't actually tell me when I need to approve CI runs, nor gives me the option to do so. Had to swap back to the old one. :sweat:

n8-dev:
Hmm, only 1 version is failing on the CI. Could it be a isolated thing?

GuySartorelli:
Looks like CI doesn't like your dist build. Did you remember to rebuild it after updating https://github.com/silverstripe/silverstripe-linkfield/pull/360#discussion_r1903386320 and https://github.com/silverstripe/silverstripe-linkfield/pull/360#discussion_r1903386519?

> only 1 version is failing on the CI

There is only one version of the JS CI job

n8-dev:
Will re-rebuild 🤷 

> > only 1 version is failing on the CI
>
> There is only one version of the JS CI job


I mean only mysql57 failed 🤔

GuySartorelli:
> I mean only mysql57 failed

Right, but it's `mysql57 js` which is the only `js` one.
The `mysql57` shouldn't *really* be there and is just a result of the way we create and name the job matrix for CI runs. 

We don't use the database for JS tests/linting, so there's no need to create another JS run with a different database, unlike unit and end-to-end tests.

n8-dev:
Woohoo that passed.

Can we merge and maybe a cheeky release @GuySartorelli ? 🎉

GuySartorelli:
Once I have tested it and approved it, I will merge it.
~~Once it has been merged, GitHub Actions will automagically tag it.~~ Sorry forgot what branch it's targetting.
It will be included as part of the April minor release.

GuySartorelli (approved):
LGTM, works well locally.
I don't understand why that client got confused, given "Cannot add links" includes the single link that field would support... but now there will be less cause for confusion.

n8-dev:
Maybe confused is not the right word.
They always want to improve CMS details for their CMS authors.

Single links were saying `links` so reported as Bug. 😄

PR #363, opened 2025-01-14
MNT Use caret for elemental dev requirement

emteknetnz:
Issue https://github.com/silverstripe/.github/issues/357

I *think* this might be preventing [linkfield merge-up workflow from working](https://github.com/silverstripe/silverstripe-linkfield/actions/runs/12761123801/job/35567645917), though maybe not.

Either way, it should be a caret requirement like it is in CMS 5

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
     },
     "require-dev": {
         "silverstripe/frameworktest": "^2",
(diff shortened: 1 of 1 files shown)


GuySartorelli (approved):
LGTM

Issue #367, opened 2025-03-05
Not all links are published when running the sheadawson/silverstripe-linkable migration task


jcop007:
### Module version(s) affected

4.1.1

### Description

We have over 1600 links on a site that we're migrating.  After running the migration task some links were not published and remain in a draft state.  Looking in the database we can see that some links had been published multiple times (had multiple versions) and others not at all.  

It appears the problem is the default sort column for the link table is "sort", which is not unique.  The link migration task chunks links into chunks of 1000.  I suspect that when it gets to the second chunk the order or results is changing and some links are being returned again and others not at all.  If I change the migration task so it returns links sorted by ID all links are published. 

### How to reproduce

In MigrationTaskTrail.php this is the relevant code:
```    /**
     * Publishes links unless Link isn't versioned or developers opt out.
     */
    private function publishLinks(): void
    {
        if (Link::has_extension(Versioned::class)) {
            $shouldPublishLinks = true;
            $i = 0;
            $this->extend('updateShouldPublishLinks', $shouldPublishLinks);
            if ($shouldPublishLinks) {
                $this->print('Publishing links.');
                /** @var Versioned&Link $link */
                **foreach (Link::get()->chunkedFetch() as $link) {
```

**To reproduce:**
Run the linkable migration task on a site which needs migrating to the new SilverStripe link module.
The default sort column for links should contain not unique values, which is expected. 
You would need to make sure the default chunk size in DataList chunkedFetch is less than the number of links to migrate. 

### Possible Solution

foreach (Link::get()->sort('ID')->chunkedFetch() as $link) {

I can submit a pull request for this if it looks like a good solution. 

### Validations

- [x] Check that there isn't already an issue that reports the same bug
- [x] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

Cambis:
I've ran into the same issue when running the gorriecoe migration task. Using the solution outlined above resolved it for me 😃

GuySartorelli:
Thanks for reporting this.

Looks like your possible solution is missing some code and an explanation? Can you please update that so it's clear what you're proposing?

jcop007:
@GuySartorelli Here is a PR for it with an explanation of the solution #370  
Thanks.

PR #370, opened 2025-03-07
FIX migration task chunking by sorting links by id.

jcop007:
## Description

 After running the linkable migration task, not all links were published.  This occurs if the number of links exceeds the default chunk limit of 1000 records. 

Looking in the database at the LinkField_Link table I could see that some links were published multiple times (2,3) and some links were not published at all, no record in the LinkField_Link_Live table.  

The default sort order for links is the "Sort" column which is not unique.  I think that when the database was queried for the second chunk the order of results was different. MySQL doesn't guarantee the same order if the sort column isn't unique.  

To fix the issue I sorted the results by ID rather than the default sort column.  This fixed the issue and all links were published once. 

## Manual testing steps
Run the link migration task against a database with over 1000 linkable links that need migrating.  Afterwards check the LinkField_Link table and the LinkField_Link_Live table to make sure that all links were published once. 

## Issues
- #367 

## Pull request checklist

- [x] All commits are relevant to the purpose of the PR (e.g. no debug statements, unrelated refactoring, or arbitrary linting)
- [x] The commit messages follow our [commit message guidelines](https://docs.silverstripe.org/en/contributing/code/#commit-messages)
- [ ] The PR follows our [contribution guidelines](https://docs.silverstripe.org/en/contributing/code/)
- [x] Code changes follow our [coding conventions](https://docs.silverstripe.org/en/contributing/coding_conventions/)
- [ ] This change is covered with tests (or tests aren't necessary for this change)
- [x] Any relevant User Help/Developer documentation is updated; for impactful changes, information is added to the changelog for the intended release
- [ ] CI is green

### DIFF
--- a/src/Tasks/MigrationTaskTrait.php
+++ b/src/Tasks/MigrationTaskTrait.php
@@ -316,7 +316,7 @@ trait MigrationTaskTrait
             if ($shouldPublishLinks) {
                 $this->print('Publishing links.');
                 /** @var Versioned&Link $link */
-                foreach (Link::get()->chunkedFetch() as $link) {
+                foreach (Link::get()->sort('ID')->chunkedFetch() as $link) {
                     // Allow developers to skip publishing each link - this allows for scenarios
                     // where links were Versioned in v2/v3 projects.
                     $shouldPublishLink = true;


GuySartorelli (approved):
LGTM, thanks

PR #372, opened 2025-04-01
API Deprecate API which was removed or changed in CMS 6

GuySartorelli:
## Issue
- https://github.com/silverstripe/.github/issues/379


src/Tasks/MigrationTaskTrait.php
```diff
@@ -22,9 +22,6 @@
 use SilverStripe\Versioned\ChangeSetItem;
 use SilverStripe\Versioned\Versioned;
 
-/**
- * @deprecated 4.0.0 Will be removed without equivalent functionality.
- */
 trait MigrationTaskTrait
```
ChloeSartorelli, line 25: We changed this so it's now deprecated as-of 5.0.0 to potentially be removed in CMS 7.

emteknetnz (approved):

PR #378, opened 2025-05-13
TLN Update translations

GuySartorelli:
Automated translations update generated using [silverstripe/tx-translator](https://github.com/silverstripe/silverstripe-tx-translator)

### DIFF
--- a/lang/en.yml
+++ b/lang/en.yml
@@ -26,7 +26,7 @@ en:
       one: 'An External Link'
       other: '{count} External Links'
     SINGULARNAME: 'External Link'
-    db_ExternalUrl: 'External URL'
+    db_ExternalUrl: 'External url'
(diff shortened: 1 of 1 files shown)


lang/en.yml
```diff
@@ -26,7 +26,7 @@ en:
       one: 'An External Link'
       other: '{count} External Links'
     SINGULARNAME: 'External Link'
-    db_ExternalUrl: 'External URL'
+    db_ExternalUrl: 'External url'
```
ChloeSartorelli, line 29: The old is more correct from a natural language POV, but must have been edited in transifex. The former is what gets actually generated. This isn't actually used as the label though (see `EXTERNAL_URL_FIELD`) so this change doesn't really matter.

GuySartorelli:
Will be regenerated

PR #380, opened 2025-05-16
TLN Update translations

GuySartorelli:
Automated translations update generated using [silverstripe/tx-translator](https://github.com/silverstripe/silverstripe-tx-translator)

### DIFF
--- a/lang/en.yml
+++ b/lang/en.yml
@@ -26,7 +26,7 @@ en:
       one: 'An External Link'
       other: '{count} External Links'
     SINGULARNAME: 'External Link'
-    db_ExternalUrl: 'External URL'
+    db_ExternalUrl: 'External url'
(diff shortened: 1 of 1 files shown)


lang/en.yml
```diff
@@ -26,7 +26,7 @@ en:
       one: 'An External Link'
       other: '{count} External Links'
     SINGULARNAME: 'External Link'
-    db_ExternalUrl: 'External URL'
+    db_ExternalUrl: 'External url'
```
ChloeSartorelli, line 29: This is the correct string collection - not sure why it was `URL` before, possibly someone changed it in transifex and it was pulled down.
Either way, this isn't even used for the FormField label so it doesn't affect anything at least used by this module directly.

emteknetnz (approved):

Issue #379, opened 2025-05-14
Change tracking bug with MultiLinkField


forsdahl:
### Module version(s) affected

4.2.1

### Description

When using the MultiLinkField, change tracking does not work as expected. When having at least one link created in a MultiLinkField, as soon as you focus any other field the change tracker will trigger and mark the form as changed, even if nothing has changed. This leads to a really annoying user experience, especially if you also happen to have an UploadField in the same form (in which case the form is marked as changed as soon as it loads).

#### Reproduction steps
1. Add a MultiLinkField to any non-react form (for example _not_ in an Elemental block)
2. Add at least one link in the field
3. Save the form
4. Close the form, and open it again
5. Focus on any field

### Possible Solution

Best way to fix this is probably to figure out what part of the react code causes this issue with the normal change tracking, however this is not immediately obvious (at least to me).

Easiest way to fix is to add the `no-change-track` extra css class to MultiLinkField, which also fixes the issue. Since the react code seems to handle change tracking directly in a slightly hackish way (adding classes to the publish button directly) I think this is a reasonale solution.

### Validations

- [x] Check that there isn't already an issue that reports the same bug
- [x] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

## PRs
- https://github.com/silverstripe/silverstripe-linkfield/pull/383

GuySartorelli:
PR merged. This will be automatically tagged by GitHub actions

PR #383, opened 2025-05-20
FIX Ensure change tracker not triggered by presence of MultiLinkField

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-linkfield/issues/379


GuySartorelli (commented):
I can't reproduce the bug without this PR which makes it impossible to test if this is actually fixing anything. I followed the reproduction steps in the issue. 

[Screencast from 21-05-25 11:02:19.webm](https://github.com/user-attachments/assets/6e5ed3d0-eea6-46fa-88c3-cc59f133abe5)

Can you please give me some steps or a recording that shows how I can reproduce the original bug?

emteknetnz:
Need to focus on a non-react field I think

I used the link field test page in framewortest and then clicked on the page title field

GuySartorelli:
That doesn't work. Can you please give me some steps or a recording that shows how I can reproduce the original bug?

GuySartorelli:
Oh, I see. The issue steps say "add a MultiLinkField to any form (for example an Elemental block)" - but apparently adding one in an elemental block _explicitly does not exhibit this problem_. I'll update the steps in the issue.

GuySartorelli (approved):
Works as expected

Issue #385, opened 2025-06-05
MultiLinkField on SiteConfig returns "Failed to load link(s)" on Save


TheSceneman:
### Module version(s) affected

4.2.3

### Description

When a MultiLinkField is added to SiteConfig via extension and `updateCMSFields()` it says "Failed to load link(s)" after the SiteConfig is saved. The links are saved, it is the retrieval of the updated list of links which is failing. See screenshot below from Network tab of Chrome dev tools. Subsequent requests add more square brackets:

![Image](https://github.com/user-attachments/assets/07b76281-d0b8-4fbe-a73d-d3110c500472)

### How to reproduce

Add extension to siteconfig
```yaml
SilverStripe\SiteConfig\SiteConfig:
  extensions:
    - App\Extensions\SiteConfigExtension
```
```php
<?php

namespace App\Extensions;

use SilverStripe\Core\Extension;
use SilverStripe\Forms\FieldList;
use SilverStripe\LinkField\Form\MultiLinkField;
use SilverStripe\LinkField\Models\Link;

class SiteConfigExtension extends Extension
{

    private static array $has_many = [
        'ConfigLinks' => Link::class . '.Owner',
    ];

    private static array $owns = [
        'ConfigLinks',
    ];

    public function updateCMSFields(FieldList $fields)
    {
        $fields->addFieldToTab('Root.Main', MultiLinkField::create('ConfigLinks'));
    }
    
}
```

Then navigate to siteconfig via `{yourdomain}/admin/settings`. Add a new link of any type via the Configlinks field.
Click save, you should see the error message. If you refresh the page you will see the links have been saved OK.

### Possible Solution

I rolled the silverstripe/linkfield version back to 4.0.0 and this fixed the issue, so the breaking change has been introduced between 4.0.0 and 4.2.3

### Validations

- [x] Check that there isn't already an issue that reports the same bug
- [x] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

## PRs
- https://github.com/silverstripe/silverstripe-linkfield/pull/386

GuySartorelli:
If the internal slack conversation is relevant, please include it here in plain text. Not everyone who can see this issue has access to our internal slack.

TheSceneman:
> If the internal slack conversation is relevant, please include it here in plain text. Not everyone who can see this issue has access to our internal slack.

[...]

PR #386, opened 2025-06-05
FIX Ensure value is in the correct format after save

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-linkfield/issues/385

### DIFF
--- a/src/Form/MultiLinkField.php
+++ b/src/Form/MultiLinkField.php
@@ -26,6 +26,14 @@ class MultiLinkField extends AbstractLinkField
         return parent::setValue($ids, $data);
     }
 
+    public function setSubmittedValue($value, $data = null)
+    {
+        if (is_string($value)) {
+            $value = rtrim(ltrim($value, '['), ']');
+        }
+        return $this->setValue($value, $data);
+    }
+
     public function getSchemaDataDefaults(): array
     {
         $data = parent::getSchemaDataDefaults();
--- a/composer.json
+++ b/composer.json
@@ -26,6 +26,7 @@
         "silverstripe/recipe-testing": "^3",
         "silverstripe/standards": "^1",
         "silverstripe/documentation-lint": "^1",
+        "silverstripe/siteconfig": "^5",
         "squizlabs/php_codesniffer": "^3",
         "phpstan/extension-installer": "^1.3"
     },
--- a/tests/behat/features/linkfield-siteconfig.feature
+++ b/tests/behat/features/linkfield-siteconfig.feature
@@ -1,0 +1,49 @@
+@retry @job2
+Feature: Create Links in LinkField and MultiLinkField on SiteConfig
+As a content editor
+I want to add links to SiteConfig which has a slightly different EditForm to SiteTree
+
+  Background:
+    Given I add an extension "SilverStripe\FrameworkTest\LinkField\Extensions\LinkPageExtension" to the "SilverStripe\SiteConfig\SiteConfig" class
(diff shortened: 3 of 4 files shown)


GuySartorelli (changes requested):
The reason `SiteConfig` does this and other admin sections don't is that `SiteConfigLeftAndMain` (and `CMSProfileController` which presumably would have the same problem) don't use the response negotiator.

i.e the problem goes away if we make this change in `SiteConfigLeftAndMain::save_siteconfig()`:

```diff
+ $this->setResponse($this->getResponseNegotiator()->respond($this->getRequest()));
  $this->response->addHeader(
      'X-Status',
      rawurlencode(_t('SilverStripe\\Admin\\LeftAndMain.SAVEDUP', 'Saved.'))
  );
- return $form->forTemplate();
+ return $this->getResponse();
```

However, we then _also_ need to update the `validationResponseCallback` defined in `getEditForm`. And that's still not the _root_ cause - there's no sensible reason why rendering the form should result in adding extra brackets around the value. It speaks to a problem in the PHP code of the form field itself.

## The root cause

The flow for most admin sections uses `getResponseNegotiator()` and renders the form fresh using `getEditForm()` rather than just rendering the form that came with the submission.

Calling `getEditForm()` fresh results in calling `$form->loadDataFrom($record);` which goes through `MultiLinkField::setValue($value, $record)` which loads the raw IDs from the relation, which _do not_ have the `'[]'` brackets around the ID array.

The flow for `SiteConfigLeftAndMain` skips loading data from the record before rendering the form, which means it has the brackets around the value, because those brackets aren't filtered out when `setSubmittedValue()` is called on the field.

## The fix

Add this method to `MultiLinkField`:

```php
public function setSubmittedValue($value, $data = null)
{
    $value = rtrim(ltrim($value, '['), ']');
    return $this->setValue($value, $data);
}
```

No need for additional javascript logic.

src/Form/MultiLinkField.php
```diff
@@ -65,7 +65,9 @@ private function convertValueToArray(mixed $value): array
         // A comma separated list of IDs will be turned into an array of IDs
         // Anything else will either get caught in the empty check or the !is_iterable check
         if (is_string($value)) {
-            $value = $this->convertCommaSeparatedString(trim($value));
+            // trim whitespace as well as any surronding brackets e.g. "[1,2,3]" => "1,2,3"
+            $value = trim($value, " \n\r\t\v\0[]");
```
emteknetnz: The 2nd param is the same default param in trim(), with the '[]' chars also included

emteknetnz:
Have updated to do it all in PHP. There was an existing method convertValueToArray() with unit tests to handle converting values so I just expanded on it

src/Form/MultiLinkField.php
ChloeSartorelli: I don't think we should be trimming all that stuff off. Only `[` and `]` are causing problems. Anything else is invalid.
Please also do this in `setSubmittedValue()` - it would be incorrect for a developer to call `$field->setValue("[1]")` but is expected for form submissions. That distinction in the API is there for this sorta thing, so we should use it.

GuySartorelli:
https://github.com/silverstripe/silverstripe-linkfield/pull/386#discussion_r2141260698 to be responded to

emteknetnz:
Updated

src/Form/MultiLinkField.php
emteknetnz: Updated

GuySartorelli (approved):
LGTM

Issue #390, opened 2025-06-23
Add key to Versioned extension on Link model


mfendeksilverstripe:
### Module version(s) affected

4.2.3

### Description

Currently, the `Versioned` extension is hard coded in the `Link` model which makes it impractical to configure.

```php
private static array $extensions = [
    Versioned::class,
];
```

### How to reproduce

Code snippet to check if Versioned extension is present.

```php
echo Link::has_extension(Versioned::class) ? 'yes' : 'no' . PHP_EOL;
```

### Possible Solution

Remove `Versioned` extension from the model.

Then either keep it like that or add it to a yaml config with a named key so it can be overridden. Depends on what the module default should be.

### Additional Context

This module should not force versioned or non-versioned links, both should be supported.

### Validations

- [x] Check that there isn't already an issue that reports the same bug
- [x] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

## PRs

- https://github.com/silverstripe/silverstripe-linkfield/pull/391

GuySartorelli:
Removing the extension by default is not an appropriate solution, but adding a key to the `extensions` config would be. That way you can just set it to `null` or `false` in project-level YAML.

Note that you can remove the extension via runtime code in `_config.php` as a workaround, which makes this a fairly low priority for the CMS Squad.
I'll happily review a PR that adds a key (e.g. `'versioned' => Versioned::class`)

mfendeksilverstripe:
Fix covered in https://github.com/silverstripe/silverstripe-linkfield/pull/391

GuySartorelli:
PR merged. Will be automatically merged by GitHub actions

PR #391, opened 2025-06-23
FIX Missing extension key.

mfendeksilverstripe:
# FIX Missing extension key.

Fix for https://github.com/silverstripe/silverstripe-linkfield/issues/390

### DIFF
--- a/docs/en/01_basic_usage.md
+++ b/docs/en/01_basic_usage.md
@@ -131,7 +131,7 @@ class MyModel extends DataObject
 
 ## Unversioned links
 
-The `Link` model has the [`Versioned`](api:SilverStripe\Versioned\Versioned) extension applied to it by default. If you wish for links to not be versioned, then remove the extension from the `Link` model via configuration `yaml` file.
+The `Link` model has the [`Versioned`](api:SilverStripe\Versioned\Versioned) extension applied to it by default. If you wish for links to not be versioned, then remove the extension from the `Link` model via YAML configuration.
 
 ```yaml
 ---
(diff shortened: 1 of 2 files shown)


GuySartorelli:
Please also update the documentation in https://github.com/silverstripe/silverstripe-linkfield/blob/5/docs/en/01_basic_usage.md#unversioned-links to suggest using YAML instead of suggesting runtime code.

mfendeksilverstripe:
I've updated the docs @GuySartorelli , please review.

docs/en/01_basic_usage.md
```diff
@@ -131,15 +131,16 @@ class MyModel extends DataObject
 
 ## Unversioned links
 
-The `Link` model has the [`Versioned`](api:SilverStripe\Versioned\Versioned) extension applied to it by default. If you wish for links to not be versioned, then remove the extension from the `Link` model in the project's `app/_config.php` file.
+The `Link` model has the [`Versioned`](api:SilverStripe\Versioned\Versioned) extension applied to it by default. If you wish for links to not be versioned, then remove the extension from the `Link` model via configuration `yaml` file.
```
ChloeSartorelli:
```suggestion
The `Link` model has the [`Versioned`](api:SilverStripe\Versioned\Versioned) extension applied to it by default. If you wish for links to not be versioned, then remove the extension from the `Link` model via YAML configuration.
```

GuySartorelli (changes requested):
Small nitpick.
Please also check the failed doc linting CI job. I think it's probably complaining that you used `yaml` instead of `yml` for the code fence

GuySartorelli (approved):
LGTM

PR #398, opened 2025-08-05
DEP Require framework ^6.1

emteknetnz:
Issue https://github.com/silverstripe/.github/issues/433

Fixes the prefer lowest build
`Error: Call to undefined method SilverStripe\ORM\DataList::reset()`
https://github.com/silverstripe/silverstripe-linkfield/actions/runs/16613554079/job/47001520014

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,7 @@
     ],
     "require": {
         "php": "^8.3",
-        "silverstripe/framework": "^6",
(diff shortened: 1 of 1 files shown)


emteknetnz:
Using https://github.com/silverstripe/silverstripe-linkfield/pull/399 instead

PR #405, opened 2025-08-21
ENH Disable sorting on queries where sort order doesn't matter

GuySartorelli:
Sorting queries can be slow, especially on large datasets and especially if indexes aren't optimised. We should avoid sorting when the sort order doesn't matter.

## Issue

- https://github.com/silverstripe/silverstripe-framework/issues/11833

### DIFF
--- a/src/Form/MultiLinkField.php
+++ b/src/Form/MultiLinkField.php
@@ -138,7 +138,7 @@ class MultiLinkField extends AbstractLinkField
         }
 
         // Load ids from relation
-        $value = array_values($relation->getIDList() ?? []);
+        $value = $relation->column('ID');
         parent::setValue($value);
     }
 
--- a/src/Tasks/MigrationTaskTrait.php
+++ b/src/Tasks/MigrationTaskTrait.php
@@ -388,7 +388,7 @@ trait MigrationTaskTrait
             foreach ($checkForBrokenLinks as $class => $data) {
                 $field = $data['field'];
                 $emptyValue = $data['emptyValue'];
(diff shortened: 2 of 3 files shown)


src/Tasks/MigrationTaskTrait.php
```diff
@@ -388,7 +388,7 @@ private function checkForBrokenLinks(): void
             foreach ($checkForBrokenLinks as $class => $data) {
                 $field = $data['field'];
                 $emptyValue = $data['emptyValue'];
-                $ids = DataObject::get($class)->filter([$field => $emptyValue])->column('ID');
+                $ids = DataObject::get($class)->filter([$field => $emptyValue])->getIDList();
```
ChloeSartorelli: Used for a count and to output the IDs that failed. Order isn't important.

emteknetnz (approved):

src/Form/MultiLinkField.php
```diff
@@ -138,7 +138,7 @@ private function loadFrom(DataObject $record): void
         }
 
         // Load ids from relation
-        $value = array_values($relation->getIDList() ?? []);
+        $value = $relation->sort(null)->column('ID');
```
emteknetnz:
```suggestion
        $value = $relation->column('ID');
```
Sort order does matter here - it's what determines the order the links are displayed on the frontend

emteknetnz (changes requested):

src/Form/MultiLinkField.php
ChloeSartorelli: Fixed and added a refresh to the behat to catch possible regressions of this in the future

emteknetnz (approved):

Issue #400, opened 2025-08-07
Accessibility: Interactive elements nested in an invalid way


GuySartorelli:
[WCAG 2.2 A 4.1.2](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html)
[ARIA 1.2 Presentational Children 5.2.9](https://www.w3.org/TR/wai-aria-1.2/#childrenArePresentational)
When an element is interactible, that marks child elements as presentational, which hides them from the accessibility tree.

When you have added a link to a link field, it is presented with a `button` like so:

`<button type='button' class='link-picker__button ...' aria-label='Edit link'>...</button>`

These contains a `<span role="button">` which is the archive/delete button. This means that the outer "edit" button will be announced, but or some screen readers even though users _can tab to_ the delete/archive button, the screen reader won't announce it.

## Possible solutions

- Use a `div` as the parent element that houses both the edit and the archive/delete button as siblings

## Acceptance criteria

- [ ] The interactive edit button doesn't contain secondary interactive elements
- [ ] The archive/delete button is a semantic `<button>` element instead of a `<span>` with button role
- [ ] The changelog clearly calls out changes to markup selectors so people can adjust their custom css/js as necessary

## PRs
- https://github.com/silverstripe/silverstripe-linkfield/pull/406
- https://github.com/silverstripe/developer-docs/pull/805

GuySartorelli:
PRs merged

PR #406, opened 2025-08-26
ENH Use button element for delete button

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-linkfield/issues/400

This PR also improves the keyboard tab order.  If starting on a MultiLinkField "Add link" button as it now goes "left to right" rather than "right to left"

Before:
- link edit button
- link archive button
- link drag handle
- next link edit button

After:
- link edit button
- link drag handle
- link archive button
- next link edit button


client/src/components/LinkPicker/LinkPickerTitle.js
```diff
@@ -170,6 +158,17 @@ const LinkPickerTitle = ({
         focusable="false"
       ></i>
     </div> }
+    {(canDelete && !readonly && !disabled) &&
+      // Intentionally using a regular button element rather than a Button react component
+      // so that we do not end up with unwanted extra css classes
+      <button
```
ChloeSartorelli, line 164: Why `button` instead of `Button`?

GuySartorelli (changes requested):
Doesn't meet the main acceptance criteria of the issue:
>  The interactive edit button doesn't contain secondary interactive elements

client/src/components/LinkPicker/LinkPickerTitle.js
emteknetnz: There's a inline comment that explains that

emteknetnz:
> Doesn't meet the main acceptance criteria of the issue:
>> The interactive edit button doesn't contain secondary interactive elements

The old element hierarchy was:

```
.link-picker__link
  .link-picker__button[aria-label="Edit link"]
    .link-picker__delete[aria-label="Archive"]
  .link-picker__drag-handle[aria-label="link-picker__drag-handle"]
```

The new element hierarchy is:

```
.link-picker__link
  .link-picker__button[aria-label="Edit link"]
  .link-picker__drag-handle[aria-label="link-picker__drag-handle"]
  .link-picker__delete[aria-label="Archive"]
```

Seems like this meets the AC?

client/src/components/LinkPicker/LinkPickerTitle.js
ChloeSartorelli: I'm blind :rofl:

GuySartorelli:
The css classes aren't helpful in seeing which elements are actually interactive.
In any case I must have been looking in the wrong place or something because I agree the two buttons are no longer nested.

GuySartorelli (approved):
LGTM

client/src/components/LinkPicker/LinkPickerTitle.js
```diff
@@ -170,6 +158,17 @@ const LinkPickerTitle = ({
...
+        aria-label={deleteText}
+        tabIndex="0"
```
ChloeSartorelli:
```suggestion
```
Oops just noticed this. We shouldn't be adding a tab index value for elements that are already in the tab order.

client/src/components/LinkPicker/LinkPickerTitle.js
```diff
@@ -170,6 +158,17 @@ const LinkPickerTitle = ({
...
+        className="link-picker__delete btn btn-link"
+        onKeyDown={handleDeleteKeyDown}
```
ChloeSartorelli, line 170:
```suggestion
```
The space and enter should fire a click event automatically for buttons.

GuySartorelli (changes requested):

client/src/components/LinkPicker/LinkPickerTitle.js
emteknetnz: It does, however if we remove this for whatever reason it seems like the dnd library gets a little confused and think that we've activated the drag handle. I don't know why, though at the same time we can just keep the code as is and things work as expected

client/src/components/LinkPicker/LinkPickerTitle.js

client/src/components/LinkPicker/LinkPickerTitle.js
ChloeSartorelli: Please add a comment to the code so when we look at this in the future we know why this seemingly redundant handler is actually necessary.

emteknetnz: Updated

GuySartorelli (approved):
LGTM

PR #414, opened 2025-10-08
ENH Hide PHP font icons from screen readers

GuySartorelli:
## Issue

- https://github.com/silverstripe/silverstripe-admin/issues/1957


GuySartorelli:
Oops incorrect branch name which is causing me problems locally.

PR #450, opened 2026-07-08
FIX Infinite re-render loop when closing "Add link" without saving a new link

Jianbinzhu:
## Description ##
LinkField's "focus after creating a new link" effect (client/src/components/LinkField/LinkField.js) never resets its focusOnNewLink state flag back to false when the user opens "Add link", picks a type (opening the create-link modal), and then closes that modal without saving.

`linkIDs` is also recomputed as a brand new array reference on every render whenever the field is empty (`if (!linkIDs) { linkIDs = []; }`). With focusOnNewLink stuck true, this effect's dependency array is considered "changed" on every single render, so it keeps re-firing indefinitely.

### Expected behaviour:
closing the "Add link" modal without saving should simply return focus to the "Add link" button and stop there.

### Observed behaviour: 
the tab freezes. I measured 100,000+ synchronous .focus() calls from a single click. The CPU spike makes every LinkField/TreeDropdownField on the page appear broken or unresponsive.

### Fix
Hoist the shared `setFocusOnNewLink(false)` call out of the if/else so it always runs

### Manual testing steps

1. Add a LinkField or MultiLinkField to any page/form.
2. Click "Add link" and choose any link type (e.g. "Page on this site") — opens the create-link modal.
3. Close the modal without saving.
4. Confirm the page remains responsive: click "Add link" again (same field or a different LinkField on the same page) and confirm it opens normally.
5. Repeat steps 2–4 a few times in a row to confirm it doesn't degrade.
6. Regression check: add a new link and save it, edit an existing link, archive/delete a link, and (for MultiLinkField) drag-reorder two or more links.
7. 
Before this fix, step 4 either fails to respond or the tab visibly freezes after step 3.

## Pull request checklist

- [ ] All commits are relevant to the purpose of the PR (e.g. no debug statements, unrelated refactoring, or arbitrary linting)
- [ ] The commit messages follow our [commit message guidelines](https://docs.silverstripe.org/en/contributing/code/#commit-messages)
- [ ] The PR follows our [contribution guidelines](https://docs.silverstripe.org/en/contributing/code/)
- [ ] Code changes follow our [coding conventions](https://docs.silverstripe.org/en/contributing/coding_conventions/)
- [ ] This change is covered with tests (or tests aren't necessary for this change)
- [ ] Any relevant User Help/Developer documentation is updated; for impactful changes, information is added to the changelog for the intended release
- [ ] CI is green


michalkleiner:
Thanks for creating the PR, @Jianbinzhu!

Can you please create an issue for this so it can be tracked properly? PRs without a corresponding issue might get lost.

Also, would you be able to add a test for this?

